PR #9, opened 2018-11-13
CMake: major refactoring. Added install command.

Thordreck:
Hi,
This pull request includes a re-factoring of the CMake. A install command has been added to be able to install the project as a standalone library (or within catkin as well) so users can just `find_package` it. The standalone installation supports version numbering. External dependencies (threads, dl and [zmq]) and headers paths are included in the cmake target, so it can be imported and used in external projects easily:

`find_package(behavior_tree_core 1.0 REQUIRED)`
`target_link_libraries(Foo behavior_tree_core)`

I had to change the minimum required CMake version up to 3.1 to use some features. Tell me if this is an issue.
Also, to be able to use it as a catkin module I had to include the 3rdparty folder in the `catkin_package`, but there was a comment explicitly discouraging it. I may have to look deeper into this.
I plan on adding deb packages generation later down the road if you think it would be a good addition.


facontidavide:
I am cherry picking your changes.

Please take a look at this branch... https://github.com/BehaviorTree/BehaviorTree.CPP/tree/cmake_fix
I have to think more about the installation.

PR #20, opened 2018-11-26
Crash fixed

Masadow:
Fixed a crash occuring when you have a Tree structure instantiated but never constructed, the destructor would then try to halt all actions but since `root_node` has never been initialized, it would just segfault.

edit: Didn't notice you used aggregate for the Tree struct, any reason you did that ? I first thought it was for some sort of C compatilibity but since the rest of the library does not use POD-classes, I guess it's not the reason


facontidavide:
Awesome!!!!
Thank you so much for the fix.

PR #27, opened 2018-12-04
Add support for ament/colcon build

mjeronimo:
Enables a colcon build for ROS2. This doesn't include porting the ROS1-specific part of xml_parsing. 

If I can get this PR integrated, I can switch to the main BehaviorTree.CPP repo, which is necessary for an upcoming release of the ROS2 Navigation stack.


facontidavide:
Thank you very much! :)

I added a little fix to fix the compilation in Indigo

mjeronimo:
Great! Thanks for the quick response!

PR #29, opened 2018-12-05
Add test dependency on ament_cmake_gtest.

nuclearsandwich:
Should resolve the failure to build on the ROS buildfarm when ament_cmake is installed but ament_cmake_gtest isn't because it isn't declared.


facontidavide:
Thanks, I am still not very familiar with ament

PR #30, opened 2018-12-05
Fix typo in package name.

nuclearsandwich:
My original PR had a typo in the package name. With https://github.com/ros2/rosdistro/pull/373 this PR should get tested on the ROS 2 buildfarm if hooks are set up (which they might not be).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,8 +44,6 @@ endif()
 find_package(ament_cmake QUIET)
 
 if ( ament_cmake_FOUND )
-  find_package(ament_cmake_gtest REQUIRED)
-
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
Still not enough to get builds passing. http://build.ros2.org/view/Cpr/job/Cpr__behaviortree_cpp__ubuntu_bionic_amd64/1/console so I'm still investigating.

nuclearsandwich:
@facontidavide in 695ad83 I removed what appears to be a spurious `find_package(ament_cmake_gtest)` and this allowed the build to pass on the ROS 2 buildfarm http://build.ros2.org/view/Cpr/job/Cpr__behaviortree_cpp__ubuntu_bionic_amd64/2/ (yay!) but it also doesn't look like there was any test output and or that tests ran.

PR #81, opened 2019-04-02
Rename tinyxml2 header so it doesn't hide the system one when installed

v-lopez:


facontidavide:
This PR fix a problem that should not be there in the first place.

tinyxml2 is supposed to be completely hidden, that is the reason why I used the Pimpl pattern.

what about removing these lines (i.e not installing anything at all) ?  https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/3rdparty/tinyXML2/CMakeLists.txt#L79-L102

v-lopez:
That would work for me also, I was going the least intrusive way, but I agree with you that they should not be installed (unless they were needed by Groot or other packages in your ecosystem).

facontidavide:
Ok, I did a change that should (finally) deal with this issue... 5815afdea3e9f339857a0b7083d97b6efb82e3f3

PR #86, opened 2019-04-05
Added return NodeStatus::RUNNING to parallel node

### LINKED ISSUE #85: Desired Behavior Parallel Node
I was using the parallel node and noticed that it shows kind of a reactive behavior. 

In the following example, `WaitForInput` is a `CoroActionNode` and will wait for a user's input, thus returning running. This will however lead to the next child of the parallel node `DoSomething` to execute.
```xml
<Parallel>
    <WaitForInput>
    <DoSomething>
</Parallel>
 ```
I could not find any detailed explanation on how the parallel node should behave in this scenario but I believe this behavior is unintuitive and doesn't fit with the semantics. 

I'm going to request a simple pull request which make its behavior more static. However, feel free to correct me if I'm missing something here.

hlzl:
Fixes #85 
With this change, the parallel node returns running as well instead of ticking its next child.

### DIFF
--- a/src/controls/parallel_node.cpp
+++ b/src/controls/parallel_node.cpp
@@ -106,8 +106,8 @@ NodeStatus ParallelNode::tick()
 
             case NodeStatus::RUNNING:
             {
-                // do nothing
-            }  break;
+                return NodeStatus::RUNNING;
+            } break;
 
             default:
             {


facontidavide:
I am sorry that I had little time to review this PR. I hope I will find time soon

hlzl:
I just had a look at this again and I think my proposal does actually not coincide with the semantics of the Parallel Node. 
I think I just got confused by the node halting running children when the threshold is reached. In the end, parallel just suggests parallel execution of the children and doesn't assure their complete execution.

facontidavide:
ok, sorry that I didn't find time to discuss this with you.

Issue #83, opened 2019-04-03
ReactiveSequence with multiple asynch children


facontidavide:
Hi @miccol 

as you know, I was unhappy with the logic of V2 about Action not being "tickable" unless set to IDLE.
Removinf that affect the ability of a ReactiveSequence (and ReactiveFallback) to have multiple asynch child (not just actions, but entire branches).

I propose this solution (separate branch):

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/reactive_stuff/src/controls/reactive_sequence.cpp

facontidavide:
Take into account that a "child" could be an AsyncAction of an entire branch that contains at least a single AsyncAction.

The idea a RUNNING state from an AsyncAction should always be propagated back to the root of the tree. Whenever a ControlNode or DecoratorNode receive a RUNNING from a child, it __must__ return RUNNING too.

For this reason, asych children should be easy to identify if RUNNING was returned at least once.

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/reactive_stuff/src/controls/reactive_sequence.cpp#L45

miccol:
Just to recap (useful for a user that is reading this for the first time and to check if we are all on the same page). Having _reactive control flow nodes_ (i.e. control flow nodes that keep re-evaluating the status of the children) with _asynchronous actions_ as children have some problems:

Consider the following BT (Fallback and Sequence are reactive. Actions are asynchronous):

![BTRaceCondition-1](https://user-images.githubusercontent.com/8132627/55487773-08a51900-562f-11e9-9b5c-bdc1207513a2.png)


Let's say that at the initial situation the Battery is OK. Hence the BT will execute the asynchronous action "Goto Kitchen", as shown below(Gray border means that the node returns _RUNNING_, Green _SUCCESS_, and Red _FAILURE_):

![BTRaceCondition-2](https://user-images.githubusercontent.com/8132627/55487791-1195ea80-562f-11e9-9a7a-53bb0f2fb0c0.png)

After some time, while the robot moves into the kitchen (e.g. Goto Kitchen is running), the battery is no longer ok. When the Fallback ticks the condition  _"Battery is OK"_, this returns _FAILURE_ and it ticks the action "_Goto Charging Station_" but... OUCH!! the action "_Goto Kitchen_" is still _RUNNING_ and we can have undetermined behaviors, as shown below :

[...]

PR #93, opened 2019-04-30
Reactive async

miccol:
Hello,

this is an attempt to solve Issue #83. 

The main important change is the implementation of the function `propagateHalt` that lets a node to trigger the ` haltChildren` routine of its parent.

@facontidavide The idea behind this is, _before_ ticking an async child, the BT should halt the other running children (if any). Let me know what do you think about this.

There are also gtests to test that tests that no race condition occurs.

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/82cb1c7b579013cc45557a00326d1b5b5d2c1cf3/tests/gtest_reactive_tree.cpp#L100-L101
 
Thank you
MC


facontidavide:
i will take my time and look at this carefully. Thanks!!!

hlzl:
Nice to see someone working on this, thanks! I'm also going to try to have a look next week.

This caught my attention:
> @facontidavide The idea behind this is, _before_ ticking an async child, the BT should halt the other running children (if any). Let me know what do you think about this.

Have you thought about letting the first async node finish execution when ticking the second async node instead of halting the first async node right away? This would mean both would run at the same time for a certain period. That's how I would imagine a reactive sequence to behave and it would allow for some nice use cases.

miccol:
> Nice to see someone working on this, thanks! I'm also going to try to have a look next week.
> 
> This caught my attention:
> 
> > @facontidavide The idea behind this is, _before_ ticking an async child, the BT should halt the other running children (if any). Let me know what do you think about this.
> 
> Have you thought about letting the first async node finish execution when ticking the second async node instead of halting the first async node right away? This would mean both would run at the same time for a certain period. That's how I would imagine a reactive sequence to behave and it would allow for some nice use cases.

That is possible already (but be careful in doing so). When sending the halt, it calls the `halt()` function, however the execution of the `tick()` is not stopped if not handled explicitly. To handle the interruption of the `tick()` execution you can do something like this:

In the `halt()` you set a variable
https://github.com/BehaviorTree/BehaviorTree.CPP/blob/68b7da74248d068a93970ab38611791282ed2797/sample_nodes/movebase_node.cpp#L36-L39


and you check it in the `tick()`
https://github.com/BehaviorTree/BehaviorTree.CPP/blob/68b7da74248d068a93970ab38611791282ed2797/sample_nodes/movebase_node.cpp#L27-L30


Anyway, If you want to discuss more about this, I think it is better if we move the conversation in another place (i.e. not in a PR).

Cheers
MC

miccol:
I think better solutions have been implemented

PR #94, opened 2019-04-30
installation of folder 3rd party done

miccol:
Hi,
this PR installs the 3rdparty folder.
Without that, the installed library cannot be used by an external project.

MC


facontidavide:
there are reasons why I don't want to install 3rdparty directory.

Which errors are you trying to fix?

miccol:
It installs only the headers.

If a project imports we get this error when making


```
 Imported target "BehaviorTree::behaviortree_cpp_v3" includes non-existent
  path

    "/home/mcolledanchise/BehaviorTree.CPP/build/install/3rdparty"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

facontidavide:
I want to fix this (without installing 3rdparty). Can you give me a way to reproduce the error?

facontidavide:
actually, check master, I think I solved the issue with the last commit

miccol:
yep!
thanks!

PR #100, opened 2019-05-24
remove redundant find_package call not conditional on BUILD_TESTING

dirk-thomas:
Follow up of #99.

The same is already present here: https://github.com/BehaviorTree/BehaviorTree.CPP/blob/1a2d1479171b46fc8e4b32fb196050261233f9ee/tests/CMakeLists.txt#L23

And the second case is also correctly guarded by a `if(BUILD_TESTING)` check which the removed duplicate lacks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,8 +51,6 @@ endif()
 find_package(ament_cmake QUIET)
 
 if ( ament_cmake_FOUND )
-    find_package(ament_cmake_gtest REQUIRED)
-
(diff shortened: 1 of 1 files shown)


dirk-thomas:
New `dev` build with this change: [Build Status](http://build.ros2.org/job/Ddev__behaviortree_cpp_v3__ubuntu_bionic_amd64/4/)

dirk-thomas:
@facontidavide The `dev` build passed. Please consider to create a new patch release from the target branch and run `bloom` for it.

PR #126, opened 2019-10-29
(3dparty coroutine) ifdef MSV_VER to WIN32

Jesus05:
On Windows not only MSVC compilator.


facontidavide:
Does this change fix any actual problem?

Jesus05:
> Does this change fix any actual problem?

Not compile with MinGW

PR #135, opened 2019-12-02
Add macOS support

3wnbr1:
Allows to compile on macOS (including latest Catalina)
- Use `ZMQ_LIBRARIES` and `ZMQ_INCLUDE_DIRS` with cmake
- define `BT_REGISTER_NODES` macro
- Replacing `std::setlocale` with `setlocale`


facontidavide:
looks good to me, thanks a lot

PR #142, opened 2020-01-07
Fixed VS2017/2019 compile error

RavenX8:
Added missing include for std::setlocale. This fixes the following error in Visual Studio:

https://ci.appveyor.com/project/facontidavide59577/behaviortree-cpp/build/job/d1ttd2w84nvnqo2e#L52


RavenX8:
Looks like MingW also has compile errors. I personally only care about VS but if you want I can take a look at those as well.

PR #152, opened 2020-03-06
Added BUILD_SHARED_LIBS option to cmake. If set to "OFF", static libr…

happykeyboard:
…ary will be generated

for a UNIX build
If BUILD_UNIT_TESTS is off, do not search for gtest library, and do not include tests subdirectory


facontidavide:
Thanks

PR #154, opened 2020-03-13
Modify documentation images.

peterpolidoro:
SequenceNode:
AimTo -> AimAt
"Aim at a target" might sound more appropriate in this example than "aim to reach a goal".

SequenceStar:
Sequence -> ReactiveSequence
The text says "On the other hand, isBatteryOK must be checked at every tick,
for this reason its parent must be a ReactiveSequence."
Modify the image to match the text.


facontidavide:
Thanks a LOT

Issue #75, opened 2019-03-25
Renaming of Sequence and Fallback node in version 3


miccol:
I think users will have a hard time with the renaming of  Sequence  to Reactive Sequence (see e.g. #74) shall we name the two nodes as Static Sequence and Reactive Sequence (or Persistent Sequence) to make the difference super clear? Unfortunately, in the BT literature, there is a bit of confusion of what a Sequence node is (if reactive or static)

facontidavide:
I try to apply the "principle of least surprise" in the API and the "static Sequence" is the one that is less surprising, therefore it should be the default one in my opinion.

Note that if you want to create an __alias name__, you can just add these two lines in your source code

```C++
BT::BehaviorTreeFactory factory;
factory.registerNodeType<FallbackNode>("StaticFallback");
factory.registerNodeType<SequenceNode>("StaticSequence");
```

miccol:
I totally agree with you about the "principle of least surprise".  I think having only Static Sequence and Reactive Sequence in the palette (i.e. not just "Sequence") will give less surprise to the users.

facontidavide:
Now that people rely on this APi, no bug changes in naming should be done.

miccol:
Actually, during the workshop many people were surprised by the unconventional naming of this library.

miccol:
@fnivek @frovida @cpaxton @doroK

facontidavide:
ok... I accept

frovida:
I agree with Michele, with "Sequence" I would expect the "ReactiveSequence" rather than the sequence with memory (or static). The majority of papers agree that reactive is the default. Having Static Sequence and Reactive Sequence in the palette seems like the safest choice.

fnivek:
Personally, I prefer the default to be the "ReactiveSequence." Behavior trees are meant to be highly reactive and the "StaticSequence" removes some of that reactivity. I want both in a library, because "StaticSequence" has important use cases. I worry that newcomers to BTs won't get the full potential of BTs because they'll just default to the static sequence. Also I suppose that "Principal of least surprise" is a relative term. I'd guess individuals from the robotics community might be more surprised by static sequence being the default than reactive sequence.

cpaxton:
I think explicitly naming each one is the safest way to go. For people who aren't versed in behavior trees, a static sequence might seem much more... sequential.

miccol:
[...]

PR #158, opened 2020-03-18
Deprecation message for SequenceNode and FallbackNode

miccol:
This PR addresses issue #75 . It is not a breaking change as the SequenceNode and FallbackNode are still there. The nodes with memory are named `LatchedSequence` and `LatchedFallback` (to be consistent with `ReactiveSequence` and `ReactiveFallback`)


facontidavide:
I am going to have a more "radical" proposal.

What if we add a parameter (input port) to Sequence and Fallback) with a boolean "isReactive"?

miccol:
or the boolean "isLatched"? I think the word "reactive" may not be intuitive for some BT users

facontidavide:
I am not planning to do this change. Thanks

PR #161, opened 2020-03-19
A couple of small changes needed to build on MSVC2015

unvestigate:
I needed these two changes to build the library on Visual C++ 2015. Using (a recent enough version of) MSVC2017 you don't need them. However, the library is quite interesting for game developers and many of us are still on old versions.


facontidavide:
Thanks a lot

PR #163, opened 2020-03-19
Disable "unreachable code" warning in xml parsing.

unvestigate:
The for loop on line 325 in xml_parsing.cpp emits a-warning-as-error "unreachable code detected", when building on MSVC in release mode. Because the body of the loop always throws an error the increment-part of the for loop is never executed. Maybe you want to change this to not use a loop in the future, but this way we can build at least. Tested on MSVC 2015.

### DIFF
--- a/src/xml_parsing.cpp
+++ b/src/xml_parsing.cpp
@@ -20,6 +20,7 @@
 
 #ifdef _MSC_VER
 #pragma warning(disable : 4996) // do not complain about sprintf
+#pragma warning(disable : 4702) // Unreachable code in "SubTree" handling.
 #endif
 
 #include "behaviortree_cpp_v3/xml_parsing.h"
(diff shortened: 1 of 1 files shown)


facontidavide:
Can you please be more specific about the warning you are trying to fix?

unvestigate:
It is warning C4702. If I try to build xml_parsing.cpp in release mode on MSVC I get this:

1>d:\projects\behaviortree.cpp\src\xml_parsing.cpp(326): error C2220: warning treated as error - no 'object' file generated
1>d:\projects\behaviortree.cpp\src\xml_parsing.cpp(326): warning C4702: unreachable code

unvestigate:
This for-loop always throws an error (both the if- and else-block). Because of this the "child = child->NextSiblingElement()" part is never executed, which MSVC does not like.

```
for (auto child = node->FirstChildElement(); child != nullptr;
                 child = child->NextSiblingElement())
            {
                if( StrEqual(child->Name(), "remap") )
                {
                   ThrowError(node->GetLineNum(), "<remap> was deprecated");
                }
                else
		{
                    ThrowError(node->GetLineNum(), "<SubTree> should not have any child");
                }
            }
```

facontidavide:
just pushed a different solution. Thanks for contributing :)

PR #176, opened 2020-04-15
Use CMake 3.5.1 as the minimum CMake version for Xenial

s-trinh:
This allows configuring on Ubuntu 16.04.

CMake 3.5.1 is the latest version in Xenial, see:

- [xenial](https://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=cmake)
- [xenial-updates](https://packages.ubuntu.com/search?suite=xenial-updates&searchon=names&keywords=cmake)


facontidavide:
thanks. I wonder ho I missed that, since I do work with Xenial ;)

Issue #92, opened 2019-04-26
Delay Decorator


ptiza-v-nebe:
Hallo!
Just as an Idea is to make counterpart to the timeout decorator that has delay functionality
It should start ticking the child after some amount of time.

facontidavide:
Feel free to send a Pull Request with this new Node

facontidavide:
I will close this soon because of inactivity

indraneelpatil:
Hello @facontidavide, since I worked on something similar, I thought maybe I could contribute this node to the library. Please review my pull request as per your schedule : https://github.com/BehaviorTree/BehaviorTree.CPP/pull/182

Also thanks for all the great work you are doing to maintain this library, it is super helpful in my work.

PR #182, opened 2020-04-26
Added delay node and wait for enter keypress node

indraneelpatil:
- This pull request is with reference to this issue [92](https://github.com/BehaviorTree/BehaviorTree.CPP/issues/92)

- Two new decorator nodes are proposed in this pull request called Delay Node and Wait for Enter Press node which I found personally very useful in my application

- The delay node allows the user to add a fixed millisecond delay before ticking the child

- The Wait for Enter Press Node as the name suggests stops execution of the child until enter key is pressed by the user

- I am very new to contributing code online and I welcome constructive criticism about this work.

I am also thankful to the owner of the library and all maintainers!


facontidavide:
First, thanks a lot for contributing.

I think the WaitForKeyPressed should use ncurses, as https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/src/controls/manual_node.cpp

facontidavide:
I see apractical problem with the DelayNode .

As you may now, Reactive nodes gives the illusion that the BT is multi-threaded, when actually it is single-threaded.

To achieve this, any code must block as little as possible. A "delay", therefore, must return RUNNING and stop execution, not freeze. 

What if the DelayNode is a child of the ReactiveSequence and it has a sibling that preempt it?

indraneelpatil:
@facontidavide Thanks for your kind words!  I had no idea Manual Selector node existed, thanks for pointing it out. Its a much better solution than what I had pushed, so I removed Wait for Key Press node in the latest commit.

indraneelpatil:
Regarding the delay node, I have added a TimerQueue timer to the node to keep a track of the delay and the node now has these three states ->

-  If the timer has not yet triggered then the node will return RUNNING
- If the timer triggers, then the node will tick the child and return the result of the child
- and if the delay node is preempted by its sibling (like in the example you specified) then the timer will be cancelled, child will not be ticked and the delay node will return failure

Please let me know if this satisfies your requirement!

indraneelpatil:
@facontidavide Hey, have you gotten any time to review this pull request? 
Thanks in advance!

include/behaviortree_cpp_v3/decorators/delay_node.h
```diff
@@ -38,18 +39,27 @@ class DelayNode : public DecoratorNode
         return {InputPort<unsigned>("delay_msec", "Tick the child after a few milliseconds")};
     }
 
+    void halt() override { 
+      
+     timer_.cancelAll();
+     DecoratorNode::halt();
+     
+      }
```
gramss, line 47: maybe remove the spaces and extra line here? So that `}` aligns with `void`

include/behaviortree_cpp_v3/decorators/delay_node.h
```diff
@@ -38,18 +39,27 @@ class DelayNode : public DecoratorNode
...
+
   private:
     TimerQueue timer_;
+     uint64_t timer_id_;
```
gramss, line 51: remove leading space?

include/behaviortree_cpp_v3/decorators/delay_node.h
```diff
@@ -30,6 +30,7 @@ class DelayNode : public DecoratorNode
 
     ~DelayNode() override
     {
+        halt();
         //timer_.cancelAll();
```
gramss, line 34: remove unused code/comment?

src/decorators/delay_node.cpp
```diff
@@ -37,19 +36,45 @@ NodeStatus DelayNode::tick()
...
+                if (!aborted)
+                {
+                    delay_complete = true;
+                }
+                else
+                {
+                    delay_aborted = true;
+                }
+            });
         }
```
gramss, line 55: add an `else` and set `delay_complete = true;` Otherwise this will delay for-ever.. If this is intended add it to your small example. Never the less, add the expected behavior for numbers <= 0 as an description ? 🙂

src/decorators/delay_node.cpp
```diff
@@ -37,19 +36,45 @@ NodeStatus DelayNode::tick()
...
     {
+        delay_complete = false;
         delay_started_ = true;
         setStatus(NodeStatus::RUNNING);
         if (msec_ > 0)
         {
-            std::lock_guard<std::timed_mutex> l(delay_mutex);
-            std::thread t(&DelayNode::delay, this);
-            t.join();
+            timer_id_ = timer_.add(std::chrono::milliseconds(msec_), [this](bool aborted) {
```
gramss, line 44: maybe add an extra line like here: https://github.com/BehaviorTree/BehaviorTree.CPP/blob/561f4fa86b45c64c9bb2fd3273f81363e6fd222f/src/decorators/timeout_node.cpp#L58
This way it gets more clear that an inline function is declared here. At least for me.. :)

include/behaviortree_cpp_v3/decorators/delay_node.h
indraneelpatil: Done

include/behaviortree_cpp_v3/decorators/delay_node.h
indraneelpatil: Done

include/behaviortree_cpp_v3/decorators/delay_node.h
indraneelpatil: Done

src/decorators/delay_node.cpp
indraneelpatil: Done

src/decorators/delay_node.cpp
indraneelpatil: Actually this is intended since we have only two cases here (I dont see any third case here which can cause the node to get stuck in delay mode) :
1.  If the timer is executed cleanly aborted flag will be false hence we will set delay_complete and the node will tick the child
2. If the timer is aborted then aborted flag will be true and we will set delay_aborted as true which will cause the delay to return failure.

indraneelpatil: In the latest commit I have handled the case of numbers<0 as a runtime error, hope this solves the query

indraneelpatil:
@gramss  Thank you for your suggestions! I have pushed them in the latest commit.

src/decorators/delay_node.cpp
```diff
@@ -53,6 +55,11 @@ NodeStatus DelayNode::tick()
                 }
             });
         }
+        else
+        {
+            throw RuntimeError("Parameter [delay_msec] in DelayNode cannot be negative (Time once lost is lost forever)! ");
```
gramss, line 60: > `(Time once lost is lost forever)!`
Couldn't have said it better.. 😄

facontidavide:
thanks and sorry for the delay. overwhelmed with other things right now

indraneelpatil:
No problem! I understand, we really appreciate all the work that you are
doing to maintain Behaviour Trees library!  I would also love to connect if
you have any future feature additions in mind and if you think I can
contribute further.

Wishing you all the health during these tough times,
Yours sincerely,
Indraneel.

PR #191, opened 2020-05-14
Fix pseudocode for ReactiveFallback.

QQting:
The pseudocode for ReactiveFallback is wrong. Index should be always reset to zero whenever it ticks again under RUNNING state. Therefore, I change it according to the source code.


facontidavide:
cool you are right. Thanks

PR #195, opened 2020-05-16
Update tutorial_01_first_tree.md

aayn:
The `Tree` class in `bt_factory.h` has no attribute called `root_node`. Most likely, it is meant to be (I tested it) `rootNode()` instead.

@facontidavide


facontidavide:
thanks a log, but the right way is tree.tickRoot();

Just updated the tutorial

PR #211, opened 2020-07-12
Update SequenceNode.md

Wuqiqi123:
Hi, thanks for your detailed document of behavior tree. I try to understand sequence but the pseudocode block my mind. Is **index** should be **_index**?


facontidavide:
thanks

PR #215, opened 2020-08-04
Allow BT factory to define clock source for TimerQueue/TimerNode

fultoncjb:
This pull request extends the utility of the TimeoutNode decorator node class to allow for the use of a custom clock. This is especially useful when someone is creating a simulated version of their code that uses a clock different from the system/steady_clock. Template parameters were added to both the TimeoutNode and TimerQueue classes so the caller may specify the clock source. Necessary changes were made in the bt_factory.cpp class to register the TimeoutNode with default template parameters. Template deduction and the associated syntax was improved in C++17, so the empty template argument list may be removed if this library is updated to C++17 in the future.


facontidavide:
Thanks a lot

PR #225, opened 2020-09-01
docs: Small changes to tutorial 02

ValerioMa:
- added implementation of Saysomething2  that was not provided
- load the tree from the defined `my_tree.xml` file


facontidavide:
Looks good to me, thanks

PR #229, opened 2020-09-12
Fix typo

doll6777:
- fix typo in SequenceNode doc (thy->they)
- fix typo in parallel_node (childred->children)


facontidavide:
thanks

PR #245, opened 2020-12-03
Update registerFromROSPlugins to support v3

amangiat88:
this allows this function to be used with v3 of the library


facontidavide:
thanks

PR #247, opened 2020-12-18
ReactiveSequence doesn't behave as it should

aayn:
Currently, the `ReactiveSequence` node doesn't seem to "RESTART" as mentioned here: https://www.behaviortree.dev/sequencenode/
Maybe there is a better way to do this (I'd be happy to update my PR based on feedback), but these changes seem to work.

@facontidavide


facontidavide:
You would break everybody code with this :( 

If a child returns FAILURE, this node must return FAILURE too

aayn:
Sorry about that. But doesn't the current `ReactiveSequence` just terminate on failure instead of restarting? @facontidavide

PR #264, opened 2021-03-04
Added printTreeRecursively overload with ostream parameter

### LINKED ISSUE #261: printTreeRecursively with custom ostream
Is there a way to call `BT::printTreeRecursively(tree.rootNode())` with custom `std::ostream`? It would be useful to use a custom `std::ostream` when a process does not use `std::cout` for logging such as for a ROS node.

asasine:
Fixes #261

### DIFF
--- a/include/behaviortree_cpp_v3/behavior_tree.h
+++ b/include/behaviortree_cpp_v3/behavior_tree.h
@@ -44,7 +44,7 @@
 #include "behaviortree_cpp_v3/decorators/timeout_node.h"
 #include "behaviortree_cpp_v3/decorators/delay_node.h"
 
-#include <ostream>
+#include <iosfwd>
 
 namespace BT
 {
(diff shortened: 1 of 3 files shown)


include/behaviortree_cpp_v3/behavior_tree.h
```diff
@@ -56,10 +57,15 @@ void applyRecursiveVisitor(const TreeNode* root_node,
...
 /**
- * Debug function to print on screen the hierarchy of the tree.
+ * Debug function to print on screen the hierarchy of the tree to std::cout.
  */
 void printTreeRecursively(const TreeNode* root_node);
 
+/**
+ * Debug function to print the hierarchy of the tree to a custom stream.
+ */
+void printTreeRecursively(const TreeNode* root_node, std::ostream& stream);
```
facontidavide, line 67: can't we just have 

> void printTreeRecursively(const TreeNode* root_node, std::ostream& stream = std::cout );

For simplicity?

facontidavide (commented):
Nice. Thanks for contributing. I added a comment

include/behaviortree_cpp_v3/behavior_tree.h
asasine: Good idea, changed it to a default argument

facontidavide:
Awesome, thanks again

PR #269, opened 2021-03-25
Clear all of blackboard's content

SteveMacenski:
Add a clearing method for all keys

Also curious, would you accept a PR to implement clearing a specific key only?


facontidavide:
Merged. Thanks.

Aboput clearing a single key, I have nothing against it, but shall we discuss the use-case?

PR #272, opened 2021-04-01
Fix bug on halt of delay node

aufheben1:
- When DelayNode is halted and ticked again, it always returned FAILURE since the state of DelayNode was not properly reset.
 - This commit fixes unexpected behavior of DelayNode when it is halted.


facontidavide:
Thanks a lot!

PR #275, opened 2021-04-23
Registered missing dummy nodes for examples

### LINKED ISSUE #259: Support inclusion or aggregation of other XML files
HI, I faced a problem in including xml files. I execute bin/t10_include_trees passing as first argument the path to examples/test_files/subtree_test.xml. Below the output:
```bash
terminate called after throwing an instance of 'BT::RuntimeError'
  what():  CheckTemperature is not a registered node, nor a Subtree
Aborted (core dumped)
```
To test I referred to the issue #17

asasine:
Example t10_include_trees was failing because nodes `CheckTemperature` and `SayHello` were not registered with the factory.

Fixes #259

### DIFF
--- a/sample_nodes/dummy_nodes.cpp
+++ b/sample_nodes/dummy_nodes.cpp
@@ -22,6 +22,12 @@ BT::NodeStatus CheckTemperature()
     return BT::NodeStatus::SUCCESS;
 }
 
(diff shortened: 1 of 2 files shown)


facontidavide:
Thanks

PR #288, opened 2021-06-10
Update fallback documentation to V3

Affonso-Gui:
Hello and as always thanks for the great work!

Updating the FallbackNode description to V3 (no more FallbackStar!) while trying to resemble at most the explanation given in [SequenceNode.md](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/docs/SequenceNode.md).

Also fixing a couple of unrelated typos in my way there.


facontidavide:
Thanks

PR #289, opened 2021-06-13
Use pedantic for non MSVC builds

ross-weir:
MSVC builds are currently failing due to:

`invalid numeric argument '/Wpedantic'`	

Only use pedantic for non MSVC builds.


facontidavide:
thanks

Issue #314, opened 2021-11-10
Build error when -DBUILD_EXAMPLES=False


Yadunund:
Greetings,

I tried building the latest `master` with `-DBUILD_EXAMPLES=False` and I encounter the error below

```
/usr/bin/ld: cannot find -lbt_sample_nodes
collect2: error: ld returned 1 exit status
make[2]: *** [tests/CMakeFiles/behaviortree_cpp_v3_test.dir/build.make:284: tests/behaviortree_cpp_v3_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:227: tests/CMakeFiles/behaviortree_cpp_v3_test.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

This happens since lib `bt_samples_nodes` is only built when `BUILD_EXAMPLES` is enabled but is linked to the [tests](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/7312ad64c2aa73e0efaa5b54d28b56d432a44194/tests/CMakeLists.txt#L31-L32) which gets [built regardless](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/7312ad64c2aa73e0efaa5b54d28b56d432a44194/CMakeLists.txt#L245-L247) 

## Instructions to reproduce error:
```bash
git clone https://github.com/BehaviorTree/BehaviorTree.CPP.git
cd BehaviorTree.CPP
mkdir build && cd build
cmake .. -DBUILD_EXAMPLES=False
make
```

asasine:
This change causes the tests to no longer be built by default. Should the default value for `BUILD_SAMPLES` be set to `ON`?

PR #315, opened 2021-11-10
Build samples independently of examples

Yadunund:
Fix https://github.com/BehaviorTree/BehaviorTree.CPP/issues/314

Signed-off-by: Yadunund <yadunund@openrobotics.org>


facontidavide:
thanks

PR #316, opened 2021-11-15
Add ENABLE_COROUTINES CMake option

fultoncjb:
Some organizations, particularly ones with strict threading models, may want to disable coroutines from the build, even if a sufficient Boost version is present on their system. This PR just exposes a new DISABLE_COROUTINES CMake option so that the user may forcefully disable them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,12 +32,6 @@ if(Boost_FOUND)
     include_directories(${Boost_INCLUDE_DIRS})
 endif()
 
-
-if(NOT DEFINED BT_COROUTINES)
-    message(STATUS "Coroutines disabled. Install Boost to enable them (version 1.59+ recommended).")
-    add_definitions(-DBT_NO_COROUTINES)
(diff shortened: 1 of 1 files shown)


facontidavide:
will accept if you set the option On by default

fultoncjb:
@facontidavide the default is OFF because the variable is in charge of disabling the coroutines (i.e. ON would mean the coroutines would be disabled). Would you like me to swap the flag convention so that it enables coroutines?

facontidavide:
Yes, don't like double negations ;)

fultoncjb:
@facontidavide fair enough. Original PR was a bit lazy to use most of the existing CMakeLists.txt layout. I have updated the convention.

facontidavide:
Thanks a lot

facontidavide:
https://github.com/BehaviorTree/BehaviorTree.CPP/runs/4923912165?check_suite_focus=true

Can you have a look at this?

PR #321, opened 2021-12-02
Specify ROS version function uses

Timple:
I was trying to get the same behavior in ROS2. But this function at least didn't do the job.


facontidavide:
Technically, there is no such thing as ROS1.

Only ROS and ROS2

Timple:
Ahwel, feel free to close if it is inconsistent with the rest of the repository.

Or merge if it clarifies things nonetheless :slightly_smiling_face:

PR #323, opened 2021-12-05
Fix Windows shared lib build

Tobias-Fischer:
This PR enables shared libs to be built on Windows as well.


facontidavide:
thanks

PR #329, opened 2022-01-19
Don't restart SequenceStar on halt

Affonso-Gui:
This might need some discussion, but I am pretty sure it's a bug.

The problem:
SequenceStar resets progress when halted https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/src/controls/sequence_star_node.cpp#L77

So in reactive formulations such as the example image below (also used in the documentation), if the battery check fails during GoTo=B, GoTo=A will be re-executed in the next sequence tick. Since we are using SequenceStar to avoid repeating actions which have already returned SUCESS, the execution should be resumed from GoTo=B even after halted.
SequenceStar

### DIFF
--- a/src/controls/sequence_star_node.cpp
+++ b/src/controls/sequence_star_node.cpp
@@ -74,7 +74,7 @@ NodeStatus SequenceStarNode::tick()
 
 void SequenceStarNode::halt()
 {
-    current_child_idx_ = 0;
+    // DO NOT reset current_child_idx_ on halt
     ControlNode::halt();
 }
 
--- a/tests/gtest_sequence.cpp
+++ b/tests/gtest_sequence.cpp
@@ -458,3 +458,89 @@ TEST_F(ComplexSequenceWithMemoryTest, Action1DoneSeq)
     ASSERT_EQ(NodeStatus::IDLE, action_1.status());
     ASSERT_EQ(NodeStatus::IDLE, action_2.status());
 }
+
+TEST_F(ComplexSequenceWithMemoryTest, Action2FailureSeq)
+{
+    root.executeTick();
+    std::this_thread::sleep_for(milliseconds(150));
+    root.executeTick();
+
+    ASSERT_EQ(NodeStatus::SUCCESS, seq_conditions.status());
+    ASSERT_EQ(NodeStatus::IDLE, condition_1.status());
+    ASSERT_EQ(NodeStatus::IDLE, condition_2.status());
+    ASSERT_EQ(NodeStatus::RUNNING, seq_actions.status());
+    ASSERT_EQ(NodeStatus::SUCCESS, action_1.status());
+    ASSERT_EQ(NodeStatus::RUNNING, action_2.status());
+
+    action_2.setExpectedResult(NodeStatus::FAILURE);
+    std::this_thread::sleep_for(milliseconds(150));
+    root.executeTick();
+
+    // failure in action_2 does not affect the state of already
+    // executed nodes (seq_conditions and action_1)
+    ASSERT_EQ(NodeStatus::FAILURE, root.status());
+    ASSERT_EQ(NodeStatus::SUCCESS, seq_conditions.status());
+    ASSERT_EQ(NodeStatus::IDLE, condition_1.status());
+    ASSERT_EQ(NodeStatus::IDLE, condition_2.status());
+    ASSERT_EQ(NodeStatus::IDLE, seq_actions.status());
+    ASSERT_EQ(NodeStatus::SUCCESS, action_1.status());
+    ASSERT_EQ(NodeStatus::IDLE, action_2.status());
+
+    action_2.setExpectedResult(NodeStatus::SUCCESS);
+    root.executeTick();
+
+    ASSERT_EQ(NodeStatus::SUCCESS, seq_conditions.status());
+    ASSERT_EQ(NodeStatus::IDLE, condition_1.status());
+    ASSERT_EQ(NodeStatus::IDLE, condition_2.status());
+    ASSERT_EQ(NodeStatus::RUNNING, seq_actions.status());
+    ASSERT_EQ(NodeStatus::SUCCESS, action_1.status());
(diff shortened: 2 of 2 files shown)


facontidavide:
thanks

Affonso-Gui:
@tgreier I think a normal Sequence node will do what you are looking for (ticking consecutive nodes without going back for as long as the node is running, and then ticking from start after the node has been reset/halted).

SteveMacenski:
@facontidavide I tend to agree with @tgreier assuming my intepretation of "halt" is correct in BT language (@miccol, thoughts?). When I call halt, I expect the behavior tree to reset all of the nodes' states, that **includes** the control flow nodes. So I think idx should be set back to zero, why would the control flow nodes be some special case relative to the other nodes? The memory of a "star" node is for intra-execution, I don't think having memory inter-execution is sensible.

SteveMacenski:
If I have a node with some `SequenceStar` branch in it to iterate over recovery behaviors -- or do any number of things as part of a task that is modeled as behavior trees (there's innumerable reasons even outside of robotics), that should be properly reset when I `halt` the tree so I can do another run of it later. I also gave a link to a specific BT in Nav2 using SequenceStar in the comment above. The memory provided by a control flow node should be during the execution of the behavior tree, not retained **between** completely different execution cycles marked by a request to reset the tree (halt). 

After this PR, it breaks that behavior by not resetting the node to be ticked next so it no longer starts with the first child like it would in a "fresh" start of the tree, which `halt` should be providing. @miccol has the same thought as me: `halt` means to reset the tree so you can have a fresh execution of that behavior tree. Do you agree with that? If so, then all of the BT nodes would need to be reset to its initial state so it can be executed again. That includes the control flow nodes states (e.g. this `current_child_idx_`). Its incredibly bizarre that only SequenceStar is the only node that doesn't have its state fully reset on a `halt` call and is asking for future bug reports / complaints. Also moreover that I think you knew this was incorrect initially when you developed it.

If you didn't want to reset the state of the BT, you should not call `halt` and instead just continue running it for your next task. I don't really see the sensibility in this PR change. 

tl;dr: I strongly encourage you to revert this PR, I don't see any reason why `halt` should not reset the state of all BT nodes. This seems like a regression and happy that The BT King, Mr. Michele Colledanchise himself agrees :laughing:

Affonso-Gui:
My main motivation for this change was:
1. To better match the (now outdated?) documentation https://github.com/BehaviorTree/BehaviorTree.CPP/blob/3.6.0/docs/SequenceNode.md#sequencestar
2. To skip having to write recognition nodes / add blackboard entries for every action, especially in early development stages. My specific use case was to look around if an object is lost or re-plan a motion if an object was moved during a pick-and-place task. I agree that adding explicit success conditions (either as recognition nodes or as state variables) would be a more "mature" implementation, but using SequenceStar helped me manage the complexity at first developments.
3. To ensure that initialization nodes are not repeated. E.g. loading the robot model, setting subscribers.

However, I do agree with the following.
> I don't think having memory inter-execution is sensible (or if there are inter-execution states, blackboard may be a good choice)

Hopefully state management has also become easier with the new version 4.0.

On the other hand, the `halt` method invocation may be a little too opaque to rely on (?). For example, in the following tree if `OtherAction` is a SyncAction that returns `SUCCESS` the SequenceStar will not be halted. If `OtherAction` is an AsyncAction, the SequenceStar will be halted as soon as the action starts (i.e. returns `RUNNING`).

test_BehaviorTree

Overall I agree with reverting this, but I would also like to have a more detailed documentation entry on this behavior.

SteveMacenski:
> (now outdated?) documentation

Yeah, I ran into alot of pages on google with deadlinks to BT.CPP's website, I'm not sure what's up with that. 

For your other points, I think those are situations perhaps that need some different designs in your BT to handle. I could give you some off the cuff suggestions, but obviously I don't know alot about your situation so they may be off base. I would strongly argue though that relying on a non-resetting SequenceStar child counter between executions is the wrong (and very fragile) thing to do.

2 suggestions though:
- Have your own `SequenceStarGlobal` BT node with "global" state stored on the blackboard that stores between runs (not best)
- Have an initialization branch that runs on a condition custom node that checks `if_initialized` on the blackboard, that is set once initialization is complete (better). That way if that branch is done, it won't be tried again, ever, even between executions. Given that the initialization bit is "task dependent" that feels very kosher and cleanly encapsulated into a single trivial, reusable node (heck, even a node that might be useful in Nav2).

> Overall I agree with reverting this

:partying_face: :fireworks:

Affonso-Gui:
Thanks for the insights. Currently I am using a version of the 1st approach (custom control node), but having a condition guard for initialization code also seems nice.

A little more clarification, though, just in case. 
> The memory provided by a control flow node should be during the execution of the behavior tree, not retained between completely different execution cycles marked by a request to reset the tree (halt).

My main use case was not to retain states across different executions, but across possible action shifts ("interruptions") within the same execution. More concretely, to avoid re-execution of completed actions when the `SEARCH` action is triggered below (i.e. implicit calls to the `halt` method).

maintree_sequencestar
findcoords_reactive


In my particular case, I would only tick the tree until the root returned `SUCCESS` or `FAILURE`. When I wanted to take another run, I would re-run the program and reconstruct a new tree, as hinted by @tgreier.

Affonso-Gui:
Taking a quick look into the Nav2 package, it seems like you are focusing more on the explicit calls to the `halt` method (https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/src/behavior_tree_engine.cpp#L86-L104), while I was more concerned with the implicit calls.
(I completely agree that it makes no sense to retain states even after explicit calls to clean up the tree).

This makes me wonder if it would be the case to simply overwride the `halt` method of the root to clean up all SequenceStar memory.
Does the example of the switching actions above classify as intra-execution (and is therefore sensible), or should it be considered as inter-execution as long as the triggered action is outside the SequenceStart childs?

SteveMacenski:
@facontidavide can this be reverted?

facontidavide:
reverted

SteveMacenski:
Thanks!

PR #332, opened 2022-01-22
Build Sample Nodes By Default to Fix Github Action

sarcasticnature:
The CMake option `BUILD_SAMPLES` was added in #315 but not enabled by default.  This causes all tests to be disabled by default, causing the github action "build and run tests" to fail


facontidavide:
thanks

PR #335, opened 2022-01-24
Fix CMake ENABLE_COROUTINES flag with Boost < 1.59

fultoncjb:


fultoncjb:
@facontidavide looks like CI build has been fixed for Boost < 1.59

PR #336, opened 2022-01-25
Dima/fix async action node

dorezyuk:
Hey,

I've noticed that the AsyncActionNode has some strange bugs in the case the tick() method throws an exception. 

1. One problem was in the thread_handle_.wait() call from within the thread. With some unlucky scheduling it might happen, that the thread_handle_ is accessed from the "tick" thread before being initialized in the "main" thread. This would lead to a (uncought) future_error exception. As a side-note bonus: This type of bugs is so prevalent, that they've made it into [this](https://www.cs.columbia.edu/~junfeng/09fa-e6998/papers/concurrency-bugs.pdf) paper (under Figure 2 from Mozilla :smile_cat:  ). Another problem is that (at least AFAIS - maybe I've missed something) there is no way to pass the wait barrier in the "tick" thread once an exception occurs.
2. Another problem is that the exptr is not cleaned up once populated - The class would keep throwing the same exception on every subsequent tick.

I've tried to add some unit tests to verify that the changes are correct. For the testing I've relied on gmock. Here I'm not entirely sure how to handle the dependencies for gmock - especially for ROS2 and non-ROS users. Maybe someone can comment on that.

Please note also that I've made a (hopefully sensible) assumption that the "tick" thread should return Failure if an exception occurs.

### DIFF
--- a/src/action_node.cpp
+++ b/src/action_node.cpp
@@ -167,6 +167,7 @@ void StatefulActionNode::halt()
 
 NodeStatus BT::AsyncActionNode::executeTick()
 {
+    using lock_type = std::unique_lock<std::mutex>;
     //send signal to other thread.
     // The other thread is in charge for changing the status
     if (status() == NodeStatus::IDLE)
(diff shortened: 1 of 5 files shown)


src/action_node.cpp
```diff
@@ -182,16 +183,23 @@ NodeStatus BT::AsyncActionNode::executeTick()
             {
                 std::cerr << "\nUncaught exception from the method tick(): ["
                           << registrationName() << "/" << name() << "]\n" << std::endl;
+                // Set the exception pointer and the status atomically.
+                lock_type l(m_);
                 exptr_ = std::current_exception();
-                thread_handle_.wait();
+                setStatus(BT::NodeStatus::FAILURE);
```
facontidavide, line 189: Does this have any effect? 
In this case we don't really want to return any status, just forward the exception

dorezyuk: My reasoning was that if the caller catchs the exception this node should still be in a "consistent" state. Without changing the state the node would be RUNNING, which to me sounds misleading. What do you think?

facontidavide: I am not sure. I would FAILURE or SUCCESS have both a semantic meaning in the tree.
I would rather set it to IDLE, to communicate that the node is not RUNNING anymore, nor returned a result, but can be restarted.

facontidavide, line 189:
```suggestion
                setStatus(BT::NodeStatus::IDLE);
```

facontidavide (changes requested):

src/action_node.cpp
dorezyuk: Will do - but I need likely to adjust the tests as well

PR #344, opened 2022-02-07
[docs] Clarify sentence

goekce:
`... will sleep up to 8 hours or less, if he/she is fully rested.` was not clear. It can also be understood as `If he/she is fully rested, the character will sleep ...`


facontidavide:
thanks

PR #346, opened 2022-02-13
Improve subtree tests

glpuga:


glpuga:
Sorry, I keep forgetting to check what repo the PR is getting created against. I would gladly share this change, but I don't think it tidy enough to bother you with it.

PR #358, opened 2022-04-07
Fixed bug where including relative paths would fail to find the correct file

### LINKED ISSUE #324: Including a tree from another directory which includes another tree fails
When you include from another directory, and that tree includes another tree, the BT factory throws the error `XML_ERROR_FILE_NOT_FOUND`.

Given these example trees and directory structure:
```
/
  tree_includes_tree_includes_same_directory.xml
  child/
    tree_includes_same_directory.xml
    tree_child_directory.xml
```

_tree_includes_tree_includes_same_directory.xml_
```xml
<root main_tree_to_execute = "IncludesTreeIncludesSameDirectory">
  <include path="child/tree_includes_same_directory.xml" />

  <BehaviorTree ID="IncludesTreeIncludesSameDirectory">
    <Sequence>
      <SubTree ID="IncudesSameDirectoryTree" />
      <AlwaysSuccess />
    </Sequence>
  </BehaviorTree>
</root>

```

_child/tree_child_directory.xml_
```xml
<root main_tree_to_execute = "ChildDirectoryTree">
  <BehaviorTree ID="ChildDirectoryTree">
    <AlwaysSuccess />
  </BehaviorTree>
</root>
```

_child/tree_includes_same_directory.xml_
```xml
<root main_tree_to_execute = "IncudesSameDirectoryTree">
  <include path="tree_child_directory.xml" />

  <BehaviorTree ID="IncudesSameDirectoryTree">
    <Sequence>
      <SubTree ID="ChildDirectoryTree" />
      <AlwaysSuccess />
    </Sequence>
  </BehaviorTree>
</root>

```

Loading the tree _tree_includes_tree_includes_same_directory.xml_ using an absolute path, it throws the above error `XML_ERROR_FILE_NOT_FOUND`.

It does not seem to matter whether the second tree (_tree_includes_same_directory.xml_ in this example) includes a third tree from the same directory, a parent directory, or a sibling directory; they all fail in the same way.

asasine:
By updating the `current_path` to the included file's absolute path before recursing, all relative paths are relative to the included file. After recursively including the new tree, the change to `current_path` is undone.

Added unit tests to verify new behavior. The new unit tests aren't built for catkin or ament because I wasn't sure where to copy the test tree files.

Because the trees are included using relative paths to the test executable, they fail if the current working directory is not the same as the _trees/_ directory i.e., `./behaviortree_cpp_v3_test` works but `./bin/behaviortree_cpp_v3_test` fails. I'd like to resolve this prior to merging but not sure how to.

Fixes #324


asasine:
I was able to make the tests pass regardless of which path the test executable is called in commit 9f9e4ab. It works by adding a [gtest environment](https://google.github.io/googletest/advanced.html#global-set-up-and-tear-down), which uses `argc` and `argv` to get the absolute path of the test executable, and making it accessible to tests through an `extern` variable.

The new unit tests still aren't built for ROS1 or ROS2. @facontidavide should I add these tests to those environments or is it fine to leave them out? I think it's just a matter of copying the trees/ directory to the proper path, but I don't have any ROS installations to test this with.

facontidavide:
I am unavailable the next week, but I will try to review and merge this after that

asasine:
@facontidavide it seems like a recent change in #373 has unintentionally broken tree inclusion. When including a file that has a tree with the `main_tree_to_execute` attribute, an exception is thrown. The most recent CI run on this branch shows the new tests I added are failing:

> C++ exception with description "The attribute [main_tree_to_execute] has been found multiple times. You must specify explicitly the name of the \<BehaviorTree\> to instantiate." thrown in the test body.

https://github.com/asasine/BehaviorTree.CPP/runs/6429134096?check_suite_focus=true#step:7:203

asasine:
I added this scenario in PR checks so the regression would be caught in the future. @facontidavide have you had a chance to review this PR and/or fix the regression?

facontidavide:
The environment trick is very neat. Thanks

facontidavide:
image

```
[ 59%] Building CXX object tests/CMakeFiles/behaviortree_cpp_v3_test.dir/gtest_ports.cpp.o
[ 60%] Building CXX object tests/CMakeFiles/behaviortree_cpp_v3_test.dir/navigation_test.cpp.o
[ 61%] Building CXX object tests/CMakeFiles/behaviortree_cpp_v3_test.dir/gtest_subtree.cpp.o
[ 63%] Building CXX object tests/CMakeFiles/behaviortree_cpp_v3_test.dir/gtest_switch.cpp.o
[ 64%] Building CXX object tests/CMakeFiles/behaviortree_cpp_v3_test.dir/gtest_coroutines.cpp.o
[ 65%] Linking CXX executable behaviortree_cpp_v3_test
Error copying directory from "/tmp/ws/src/behaviortree_cpp/tests/trees" to "/trees".
gmake[2]: *** [tests/CMakeFiles/behaviortree_cpp_v3_test.dir/build.make:327: tests/behaviortree_cpp_v3_test] Error 1
gmake[2]: *** Deleting file 'tests/behaviortree_cpp_v3_test'
gmake[1]: *** [CMakeFiles/Makefile2:373: tests/CMakeFiles/behaviortree_cpp_v3_test.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
---
--- stderr: behaviortree_cpp_v3
CMake Warning:
  Manually-specified variables were not used by the project:

    CATKIN_INSTALL_INTO_PREFIX_ROOT
    CATKIN_TEST_RESULTS_DIR


Error copying directory from "/tmp/ws/src/behaviortree_cpp/tests/trees" to "/trees".
gmake[2]: *** [tests/CMakeFiles/behaviortree_cpp_v3_test.dir/build.make:327: tests/behaviortree_cpp_v3_test] Error 1
gmake[2]: *** Deleting file 'tests/behaviortree_cpp_v3_test'
gmake[1]: *** [CMakeFiles/Makefile2:373: tests/CMakeFiles/behaviortree_cpp_v3_test.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
---
Failed   <<< behaviortree_cpp_v3 [56.3s, exited with code 2]
```

asasine:
The broken build appears to be from a missing `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}` variable

facontidavide:
Atually:

> Error copying directory from "/tmp/ws/src/behaviortree_cpp/tests/trees" to "/trees".

This is happening on the Open Robotics CI server. Not sure how to set those variables

asasine:
It’s set by the root CMakeLists.txt

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/c14f25c3d4aa09c19296cd6bfdf0b1bb3ba9c37b/CMakeLists.txt#L255

asasine:
@facontidavide I believe this is failing because the Open Robotics CI server is building with ament, `BUILD_TESTING` might be off, and `BUILD_UNIT_TESTS` is on by default. Therefore, the ROS2/ament branch of tests/CMakeLists.txt would be skipped, and the pure-CMake block in tests/CMakeLists.txt is executed.

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/4fa2177850cb0690e46ebb388f1db7293548bb95/tests/CMakeLists.txt#L53-L69

Since `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is never set, `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/trees` evaluates to `/trees`, which cmake would not have valid permissions to write to.

My recommendation is to remove `BUILD_UNIT_TESTS` in favor of the canonical `BUILD_TESTING` flag instead. Alternatively, defaulting `BUILD_UNIT_TESTS` to the value of `BUILD_TESTING` might work too.

facontidavide:
I think the problem is not if the test is build or not. If it is built even if BUILD_TESTING is not set, no bug deal.

The problem is really 
`set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/${BEHAVIOR_TREE_BIN_DESTINATION}" )`

And the fact that **CMAKE_BINARY_DIR** is not set.

Have a look at this, I think it is a better way to address the issue: https://github.com/BehaviorTree/BehaviorTree.CPP/pull/398

asasine:
Removing the use of `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is also good as it's extra bloat but I don't think it's the root problem.

It's safe to rely on `CMAKE_BINARY_DIR` being set by CMake https://cmake.org/cmake/help/v3.10/variable/CMAKE_BINARY_DIR.html.  A bigger issue (imo) is that tests are being built when the caller is requesting them not to be through the canonical flag `BUILD_TESTING=OFF`, and the code in _tests/CMakeLists.txt_ for building pure-CMake tests assumes the same corresponding block in the top-level _CMakeLists.txt_ is also executed, which it is not.

---

This entire block

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/4fa2177850cb0690e46ebb388f1db7293548bb95/CMakeLists.txt#L250-L257

isn't even executed in the Open Robotics CI build. That build would be calling this branch instead

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/4fa2177850cb0690e46ebb388f1db7293548bb95/CMakeLists.txt#L234-L245

since `ament_cmake_FOUND` would be true.

The bug in the Open Robotics CI build only crops up because this branch of _test/CMakeLists.txt_ 

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/4fa2177850cb0690e46ebb388f1db7293548bb95/tests/CMakeLists.txt#L53-L67

assumes that the pure CMake branch of the top-level _CMakeLists.txt_ was executed which sets `CMAKE_RUNTIME_OUTPUT_DIRECTORY` but that's branch isn't executed, the `ament_cmake_FOUND` one is. ament_cmake branch in the top-level _CMakeLists.txt_, pure CMake in the _tests/CMakeLists.txt_.

---

Another option is to restructure the branch in _tests/CMakeLists.txt_ to look like this

```cmake
if(ament_cmake_FOUND)
  if(BUILD_TESTING)
    # ...
  endif()
elseif(catkin_FOUND)
  if(CATKIN_ENABLE_TESTING)
    # ...
  endif()
else()
  if(BUILD_UNIT_TESTS)
    # ...
  endif()
endif()
```

or restructure the `add_subdirectory(tests)` to look like this

```cmake
if(BUILD_SAMPLES AND ((ament_cmake_FOUND AND BUILD_TESTING) OR (catkin_FOUND AND CATKIN_ENABLE_TESTING) OR (NOT ament_cmake_FOUND AND NOT catkin_FOUND AND BUILD_UNIT_TESTS)))
  add_subdirectory(tests)
endif()
```

PR #362, opened 2022-04-15
Export dependency on ament_index_cpp

hsmin99:
To make dependent packages try to link ament_index_cpp, export the
dependency explicitly.


facontidavide:
thanks

PR #363, opened 2022-04-15
threshold child count dynamically in parallel control node

zflat:
This is a modification to the `ParallelNode` control to allow negative success threshold and failure threshold values that act like the negative index in python arrays to count backwards from the child node count. This allows for thresholds to be dynamic so that additions or removals of child nodes in an XML file will not require an update to thresholds if the intention is to have threshold values based on the total number of children.


facontidavide:
thanks, it is a good change

PR #368, opened 2022-04-27
Change order of lock to prevent deadlock.

### LINKED ISSUE #367: [Blackboard] Deadlock on concurrent set() and getInput().
Hello @facontidavide,

We found the following deadlock using `behaviortree_cpp` version 3.6.1.
It deadlocks in the following case using the newly introduced `entry_mutex_` from https://github.com/BehaviorTree/BehaviorTree.CPP/commit/06a1f4cedd6192af9d7e2a9a4b0da3d369a02c8d. 

Lets think about the following two concurrent threads:

Thread A calls `Blackboard::set()`
Thread A locks `mutex_`
Thread B calls `TreeNode::getInput()`
Thread B locks `entry_mutex_`
Thread A waits for `entry_mutex_`
Thread B waits for `mutex_`

The order of locking introduces the deadlock.

Mitigation:
In `Blackboard::set()` first lock `entry_mutex_` then lock `mutex_`.

Thread A calls `Blackboard::set()`
Thread A locks `entry_mutex_`
Thread B calls `TreeNode::getInput()`
Thread B waits for `entry_mutex_`
Thread A locks `mutex_`
Thread A releases `mutex_` and `entry_mutex_`
Thread B locks `entry_mutex_`
Thread B locks `mutex_`
Thread B releases `mutex_` and `entry_mutex_`

Thanks in advance!

FabianSchurig:
Resolves #367.

### DIFF
--- a/include/behaviortree_cpp_v3/blackboard.h
+++ b/include/behaviortree_cpp_v3/blackboard.h
@@ -116,8 +116,8 @@ class Blackboard
     template <typename T>
     void set(const std::string& key, const T& value)
     {
-        std::unique_lock<std::mutex> lock(mutex_);
         std::unique_lock<std::mutex> lock_entry(entry_mutex_);
+        std::unique_lock<std::mutex> lock(mutex_);
         auto it = storage_.find(key);
 
         if( auto parent = parent_bb_.lock())


facontidavide:
I should probably add a unit test for this....

PR #379, opened 2022-05-14
Revisit export configuration

artivis:
While building ROS 2 from source together with Nav2 and BT I've encountered a couple issues with BT.

- First the `Config.cmake` file exported was named `BehaviorTreeV3Config.cmake` whereas, the project being named `behaviortree_cpp_v3`, one would expect the file to be `behaviortree_cpp_v3Config.cmake`.
- Second, for some reason I would never enter the `if(ament_cmake_FOUND)` conditional block, thus the `behaviortree_cpp_v3_TARGETS` would never be automatically exported by `ament` and thus never found when BT is included in another project; `ament_target_dependencies` depends on it to properly set up the interfaces.

This PR adresses both issues by revisiting the project export configuration following the [tutorial on CMake website](https://cmake.org/cmake/help/latest/guide/tutorial/Adding%20Export%20Configuration.html).

Marking as draft since I've only tested my use case and this may break others.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -279,7 +279,7 @@ endif()
 # INSTALL
 
 INSTALL(TARGETS ${BEHAVIOR_TREE_LIBRARY}
-    EXPORT BehaviorTreeV3Config
+    EXPORT ${PROJECT_NAME}Targets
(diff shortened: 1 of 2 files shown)


facontidavide:
Look good to me, but did you check if the NON ROS compilation still work?

Issue #388, opened 2022-05-23
Last commit broke importing with CMake


DatSpace:
The last commit from merged from the PR #379 seems to have broken cmake from being able to find the file. The name of the generated file has been changed and thus doing find_package does not find the configuration.

To replicate pull and compile this repo and create another project and try to find it. In my case I do not install, but instead specify the path to the build directory by setting the env var.

PR #389, opened 2022-05-24
Fix destination in CMakeLists.txt

galou:
Signed-off-by: Gaël Écorchard <gael.ecorchard@cvut.cz>

I'm not a CMake expert but this solved #388 on my setup.


facontidavide:
thanks

PR #404, opened 2022-07-20
Check for / install GTest even if using catkin or ament

paulerikf:
I could very well be missing something here, but creating a PR just in case.

Builds with colcon were failing for me with `gtest/gtest.h file not found`

Previously gtest was checked for / installed in an `elseif(BUILD_UNIT_TESTS)` block, which would only trigger if neither ament or catkin were found.

All this PR does is change the `elseif(BUILD_UNIT_TESTS)` to an `if(BUILD_UNIT_TESTS)`, which fixed my build issues.

@facontidavide Does this look ok? Or am I missing something else?


facontidavide:
change merged, thanks!

facontidavide:
reverted. It breaks compilation in ROS

paulerikf:
Ah, in ROS1? I'll try to find something that works for both.

PR #421, opened 2022-09-01
Add vcpkg installation instructions

FrankXie05:
`BehaviorTree.CPP` is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `BehaviorTree.CPP` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `BehaviorTree.CPP`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/behaviortree-cpp/portfile.cmake). We try to keep the library maintained as close as possible to the original library.😊

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,18 @@ You can easily install the package with the command
 If you want to compile it with catkin, you __must__ include this package 
 to your catkin workspace.
 
+## Installing BehaviorTree.CPP (vcpkg)
+
+Alternatively, you can build and install behaviortree-cpp using [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install behaviortree-cpp
+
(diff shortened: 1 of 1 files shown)


antonkesy (approved):

facontidavide:
I do not maintain vcpkg and I don't know who does.
Apparently the latest supported version is 4.1.1 that I DO NOT want to encourage people to use the old and buggy version.

FrankXie05:
@facontidavide I will update this.

FrankXie05:
The version on vcpkg has been update.

facontidavide:
I bug problem I see there is this:

        -DBTCPP_GROOT_INTERFACE=OFF
        -DBTCPP_SQLITE_LOGGING=OFF
        
I know this makes it easier, but loosing those dependency is a big deal, in my opinion.

FrankXie05:
These dependencies are mainly used for the construction of `conan`. I can solve it through patches, but this method only applies to vcpkg. I would like to ask if this method is allowed? 🫤

BTW, we may not maintain the conan problem, and my patch is just to change the dependency to vcpkg usage.

Here is the patch:
[fix-conan.patch](https://github.com/BehaviorTree/BehaviorTree.CPP/files/12746217/fix-conan.patch)

PR #438, opened 2022-09-28
Fix incorrect registration of behavior trees containing faulty XML

schornakj:
This fixes a problem where a behavior tree containing an invalid structure (for example, a BehaviorTree with less than one child or more than two children) could still be instantiated by the factory, even though the XML parser threw and exception when trying to load it. This was possible because adding a BehaviorTree's name to the list of available behavior trees occurred before the check for the validity of the BehaviorTree's XML.

- Modify XML parsing to add a BehaviorTree node to the list of available trees only if that BehaviorTree is valid.
- Delete unused `tree_names` and `tree_count` variables.
- Extend the unit tests to reflect this error case. These tests fail without the fix in this PR.
- ~~Move XML parsing helper functions into an anonymous namespace so they can be used in more places.~~

### DIFF
--- a/include/behaviortree_cpp_v3/xml_parsing.h
+++ b/include/behaviortree_cpp_v3/xml_parsing.h
@@ -36,6 +36,9 @@ private:
   Pimpl* _p;
 };
 
+void VerifyXML(const std::string& xml_text,
+               const std::unordered_map<std::string, NodeType>& registered_nodes);
+
 std::string writeTreeNodesModelXML(const BehaviorTreeFactory& factory,
                                    bool include_builtin = false);
 
--- a/include/behaviortree_cpp_v3/xml_parsing.h
+++ b/include/behaviortree_cpp_v3/xml_parsing.h
@@ -36,9 +36,6 @@ private:
   Pimpl* _p;
 };
 
-void VerifyXML(const std::string& xml_text,
-               const std::unordered_map<std::string, NodeType>& registered_nodes);
-
 std::string writeTreeNodesModelXML(const BehaviorTreeFactory& factory,
                                    bool include_builtin = false);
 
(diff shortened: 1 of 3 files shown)


include/behaviortree_cpp_v3/xml_parsing.h
```diff
@@ -36,9 +36,6 @@ class XMLParser : public Parser
   Pimpl* _p;
 };
 
-void VerifyXML(const std::string& xml_text,
```
facontidavide, line 39: I think you should not remove this from the public interface

facontidavide:
Thanks for the help.
But I think that you should not remove the function from having public access, since Groot 1.0 relies on that.

Also, seems like you moved some code without much modification, and this makes it very hard for me to understand if you changed something or not.

src/xml_parsing.cpp
```diff
@@ -46,6 +46,188 @@ auto StrEqual = [](const char* str1, const char* str2) -> bool {
...
+    count++;
+  }
+  return count;
+}
+
+/**
+ * @brief Recursively enforces that the attributes of an XML node and all its child nodes are all correct.
+ * @throw BT::RuntimeError if an error was found.
+ */
+void verifyXMLRecursively(
```
facontidavide, line 79: DO not move this code (hard for me to compare) and keep name and namespace

facontidavide (changes requested):

include/behaviortree_cpp_v3/xml_parsing.h
schornakj: Understood, I put it back. Good to know that Groot uses this too.

src/xml_parsing.cpp
schornakj: I reverted my changes to these functions to keep the inline helper functions, which reduces the size of the diff.

PR #439, opened 2022-09-30
Add functionality to clear registered behavior trees

schornakj:
Add functions to `BehaviorTreeFactory`, the base `Parser` class, and `XMLParsing` to clear the parser's internal list of registered behavior trees, plus supporting unit tests.

This is needed for cases where the factory is loading user-defined behavior trees during runtime, as opposed to just loading them once at startup, since if the  user deletes or edits the behavior tree definitions the factory needs to be able to reload them from scratch.

Based on work by @jliukkonen.

### DIFF
--- a/include/behaviortree_cpp_v3/bt_factory.h
+++ b/include/behaviortree_cpp_v3/bt_factory.h
@@ -317,6 +317,11 @@ public:
   std::vector<std::string> registeredBehaviorTrees() const;
 
   /**
+   * @brief Clear previously-registered behavior trees.
+   */
+  void clearRegisteredBehaviorTrees();
+
+  /**
      * @brief instantiateTreeNode creates an instance of a previously registered TreeNode.
      *
      * @param name     name of this particular instance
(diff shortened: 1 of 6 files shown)


facontidavide:
I am not sure that you really need this.

If you want to clean an entire Parser or factory, why not:

```c++
// first creation
BehaviorTreeFactory factory;
// clean up
factory = BehaviorTreeFactory{};
```
The effect should be exactly the same;

schornakj:
> I am not sure that you really need this.
> 
> If you want to clean an entire Parser or factory, why not:
> 
> // first creation
> BehaviorTreeFactory factory;
> // clean up
> factory = BehaviorTreeFactory{};
> 
> The effect should be exactly the same;

Deleting and recreating the whole BehaviorTreeFactory has the additional effect of de-registering all the node builder functions which have been registered up to that point. I think it's beneficial to allow clearing and reloading the registered behavior trees without requiring that custom builders be reloaded. Depending on the implementation of the class owning the BehaviorTreeFactory, there could be overhead related to registering the custom builders such as loading plugins from pluginlib which would be undesirable to repeat unnecessarily.

Behavior tree XML is simple to implement and load on the fly while a particular instance of the BehaviorTreeFactory remains running. Adding a new behavior tree node implementation requires modifying C++ code and either changing the code of the class that contains the BehaviorTreeFactory or registering a new plugin. I think it'll be much more common for someone to want to do runtime-loading of a new behavior tree XML definition vs. a new node type -- the first case is way easier to enable using tools already provided though BT.Cpp anyway.

PR #426, opened 2022-09-13
Fix: PublisherZMQ::flush is called after the publisher has been destructed

Tradias:
There are unfortunately no tests for any of the loggers :(. But adding them seemed out-of-scope for this pr.

Test code:

```cpp
    static constexpr auto XML = R"(
<root>
    <BehaviorTree>
    <SetBlackboard output_key="arg1" value="1" />
    </BehaviorTree>
</root>
)";
    BT::BehaviorTreeFactory factory;
    auto tree = factory.createTreeFromText(XML);
    BT::PublisherZMQ zmq_logger{tree, 1};
    tree.tickRoot();
```

Output:

```
Exception thrown at 0x00007FFE533E4FD9 in test-behavior-tree.exe: Microsoft C++ exception: zmq::error_t at memory location 0x0000005A1F7AEB40.
[PublisherZMQ] Publisher quitting.
[PublisherZMQ] just died. Exception Context was terminated
Exception thrown at 0x00007FF659E7124F in test-behavior-tree.exe: 0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF.
```

PR #440, opened 2022-09-30
Fix PublisherZMQ segfault upon destruction

schornakj:
Wait for any in-progress message sending to finish before deleting ZMQ. This fixes some cases where a segfault could occur when an instance of PublisherZMQ was destroyed.

 I added unit tests based on the test case described in #426 (side note: that PR has a somewhat more complex solution to this same problem, but I pushed this one because I think the solution here is OK). I also added the bonus esoteric-yet-seen-in-the-wild case of deleting PublisherZMQ without even ticking the tree.

This is based on work by @JafarAbdi.

### DIFF
--- a/src/loggers/bt_zmq_publisher.cpp
+++ b/src/loggers/bt_zmq_publisher.cpp
@@ -92,6 +92,10 @@ PublisherZMQ::~PublisherZMQ()
   }
   flush();
   zmq_->context.shutdown();
+  if (send_future_.valid())
+  {
+    send_future_.wait();
+  }
   delete zmq_;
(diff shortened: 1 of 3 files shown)


Tradias:
The reason I didn't go for this simple implementation is that the destructor will now block for up to `min_time_between_msgs_`, in my implementation it would complete "immediately".

Tradias:
Aside from the fact that `PublisherZMQ::callback` still has a race condition on `send_pending_`.

facontidavide:
ok, thanks, I will reconsider and merge yours

PR #455, opened 2022-10-26
Feature: migration script for V3 xml files to V4 format

SubaruArai:


SubaruArai:
I'm not sure how codeql is configured, but it seems more like a configuration problem than my PR's problem.

facontidavide:
I think this is really really great. I am alittle busy right no to review it, but I will do it soon. Thanks

facontidavide:
thanks a lot, I just changes a couple of things

PR #462, opened 2022-11-03
bugfix: license for filesystem library

SubaruArai:
The `filesystem` in `3rdparty` didn't have its legal license.


facontidavide:
this entire library should be removed, since I am using the c++17 version

facontidavide:
entire dependency removed

PR #472, opened 2022-11-14
Add the RunOnce decorator node

galou:
This adds a RunOnce decorator node that will call its child' `executeTick()` only once. After the first time, it'll return the return value of child's `status()`.

Signed-off-by: Gaël Écorchard <gael.ecorchard@cvut.cz>

### DIFF
--- a/include/behaviortree_cpp/decorators/run_once_node.h
+++ b/include/behaviortree_cpp/decorators/run_once_node.h
@@ -1,0 +1,54 @@
+/*  Copyright (C) 2018-2020 Davide Faconti, Eurecat -  All Rights Reserved
+*   Copyright (C) 2022 Gaël Écorchard, Czech Institute of Informatics, Robotics, and Cybernetics (ciirc) <gael.ecorchard@cvut.cz>
+*
+*   Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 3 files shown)


facontidavide:
It is ok for me to merge this

include/behaviortree_cpp/decorators/run_once_node.h
```diff
@@ -0,0 +1,54 @@
...
+
+  if (already_ticked_)
+  {
+    return child_node_->status();
+  }
+  else
+  {
+    return child_node_->executeTick();
+  }
+  already_ticked_ = true;
```
facontidavide, line 52: I think this is what you want:

```suggestion
  else
  {
    already_ticked_ = true;
    return child_node_->executeTick();
  }
```

galou:
I changed according to your wish but the functionality was not affected.

facontidavide:
I am still thinking about what the desired behavior is for Asynchronous children.

May be this make more sense?

```c++
auto status = child_node_->executeTick();
if(status == NosteStatus::SUCCESS || status == NosteStatus::FAILURE) {
   already_ticked_ = true;
}
return status;
```

Otherwise we don't complete RUNNING children

facontidavide:
no, not necessarily

facontidavide:
Merged with some important improvements 

3efde1e795749f4d1d982ed3dc53c3f2745e39e9

galou:
Nice, sorry for not having reacted on this.

PR #477, opened 2022-11-24
Define NodeConfiguration for BT3 compatibility

galou:
Define `NodeConfiguration` for BT3 compatibility

I didn't compile because I don't need BT3 compatibility.


facontidavide:
thanks!

Issue #489, opened 2023-01-02
Port type missmatch between blackborad and subtree port


farshbafdoustar:
I have an ActionNode (Action1) that has input port with std::vector<int> type:

```
<BehaviorTree ID="Main">
   <SubTree ID="Subtree_A" states="1;2"/>
  </BehaviorTree>

<BehaviorTree ID="Subtree_A">
    <Sequence>
        <Action1 states="{states}"/>
    </Sequence>
  </BehaviorTree>
```
When I want to pass sth to this port through a wrapper Subtree, I got the following error:

```The creation of the tree failed because the port [states] was initially created with type [std::string] and, later type [std::vector<int, std::allocator<int> >] was used somewhere else.```

it seems that the blackboard item defined in Main tree is created by default with std::string type. Is there anyway to define the type for the blackboard items in advance or any workaround to handle such a situation?

facontidavide:
Let me see if I can replicate this issue locally

facontidavide:
is this the master branch (version 4.0.1?)

facontidavide:
fix pushed to master. thanks for reporting

farshbafdoustar:
Wow, Thanks Davide!

PR #490, opened 2023-01-02
fixing missed return type related to issue 489

farshbafdoustar:
fixing missed return type error occurred during build after commit 5e16d72c18a4990a73e2a3cdef9150805e4a66ea related to [issue 489](https://github.com/BehaviorTree/BehaviorTree.CPP/issues/489#issue-1516359545)


facontidavide:
thanks, but I fixed it already!

PR #497, opened 2023-01-19
using tinyxml ErrorStr() instead of ErrorName() to get more info about missing xml file

farshbafdoustar:
During the loading process of BT from an XML file that includes other xml files, inclusion of a not existing file (or misspelled name/path) leads to the following error on the console:
```
Error parsing the XML: XML_ERROR_FILE_NOT_FOUND
```
but this give no info about which file is missing .
using doc->ErrorStr() instead of doc->ErrorName() will result in :
```
Error parsing the XML: Error=XML_ERROR_FILE_NOT_FOUND ErrorID=3 (0x3) Line number=0: filename=/[path_to_file]/[filename].xml
```


facontidavide:
thanks

PR #518, opened 2023-03-10
fix alloc-dealloc-mismatch for _storage.str.data

alsora:
The variable `_storage.str.data` is a `char*` allocated as

```
_storage.str.data = new char[size + 1];
```

This means that we should use `delete[]` to deallocate it, to avoid undefined behavior.

### DIFF
--- a/include/behaviortree_cpp/utils/simple_string.hpp
+++ b/include/behaviortree_cpp/utils/simple_string.hpp
@@ -53,7 +53,7 @@ class SimpleString {
   ~SimpleString()
   {
     if (!isSOO()) {
-      delete _storage.str.data;
+      delete[] _storage.str.data;
     }
     _storage.soo.capacity_left = CAPACITY;
   }


alsora:
One of the jobs failed with the following error which seems unrelated
```
  Errors     << behaviortree_cpp:make /root/target_ws/logs/behaviortree_cpp/test.make.000.log
  /usr/bin/ld: /usr/lib/x86_64-linux-gnu/libgtest.a(gtest-all.cc.o): undefined reference to symbol 'pthread_getspecific@@GLIBC_2.2.5'
  /usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
  collect2: error: ld returned 1 exit status
  make[3]: *** [tests/CMakeFiles/behaviortree_cpp_test.dir/build.make:350: /root/target_ws/devel/.private/behaviortree_cpp/lib/behaviortree_cpp/behaviortree_cpp_test] Error 1
  make[2]: *** [CMakeFiles/Makefile2:742: tests/CMakeFiles/behaviortree_cpp_test.dir/all] Error 2
  make[1]: *** [CMakeFiles/Makefile2:476: CMakeFiles/run_tests.dir/rule] Error 2
  make: *** [Makefile:253: run_tests] Error 2
  cd /root/target_ws/build/behaviortree_cpp; catkin test --get-env behaviortree_cpp | catkin env -si  /usr/bin/make run_tests; cd -

```

facontidavide:
thanks

Issue #525, opened 2023-03-14
Pre-Condition : _skipIf NOT working with ReactiveSequence


ephson24:
Hi Davide,

I have this simple behavior to check battery level(condition always returning TRUE) and only send warning only if the "isLowBattery" variable is set. Thus, the **SendWarning** node should not run if it is false.
`<root BTCPP_format="4" >
  <BehaviorTree ID="PowerManagerT">
    <ReactiveSequence>
      <Script code=" LOW_BATT:=20 "/>
      <CheckLevel deviceType="BATT"
                  percentage="{LOW_BATT}"
                  isLowBattery="{isLowBattery}"/>
      <SendWarning deviceType="BATT" deviceStatus="{isLowBattery}" _skipIf="!isLowBattery"/>
    </ReactiveSequence>
  </BehaviorTree>
</root>`

But when running the tree, **SendWarning** keeps getting called regardless. However, if I change the control node to use a **Sequence**, it works as expected.

PS: I am using the same cpp code refrenced in [this issue](https://github.com/BehaviorTree/BehaviorTree.CPP/issues/523#issue-1621520744)
With the added code fixes when using **getInput**.

facontidavide:
Please look at this. it seems to work correctly to me:

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/tests/script_parser_test.cpp#L280-L288

Also, there was an error in your code that I corrected:

``` c++
 ((100 < percent)? true : false))
``` 
This will return false.
I changed it to: 

``` c++
 ( percent < 25 )
```

facontidavide:
Closing this. If you think that my unit test does not reproduce correctly your issue, please suggest a modification to the unit test and send a PR.

facontidavide:
I was able to reproduce the issue with much simpler code. Thanks a lot for your help!!!

https://github.com/BehaviorTree/BehaviorTree.CPP/commit/c71b2aaa759230d755d7d7ba1ec8056606e82565

facontidavide:
problem found, I think, it is a bug in **SimpleActionNode**.

I will keep you posted

facontidavide:
better unit test added. Have a look

ephson24:
> better unit test added. Have a look

Will do. Thanks

PR #526, opened 2023-03-14
 ReactiveSequence _skipIf issue fix

ephson24:
Hi Davide, 

I fixed the _skipIf issue described in [this issue](https://github.com/BehaviorTree/BehaviorTree.CPP/issues/525#issue-1623185459) and created two tests: **SkippingLogic.RepeatedSkippingReactiveSequence** and **SkippingLogic.RepeatedNoSkippingReactiveSequence** to verify.

**PS**: The issue is only observable when ticking the tree continuously. E.g Calling tickOnce() inside a while loop. I hope this helps.

### DIFF
--- a/src/controls/reactive_sequence.cpp
+++ b/src/controls/reactive_sequence.cpp
@@ -60,7 +60,7 @@ NodeStatus ReactiveSequence::tick()
     }   // end switch
   }     //end for
 
-  if (success_count == childrenCount())
+  //if (success_count == childrenCount())
   {
     resetChildren();
   }
--- a/tests/gtest_skipping.cpp
+++ b/tests/gtest_skipping.cpp
@@ -3,6 +3,7 @@
 #include "behaviortree_cpp/basic_types.h"
 #include "behaviortree_cpp/bt_factory.h"
 #include "test_helper.hpp"
+#include "../sample_nodes/dummy_nodes.h"
 
 using namespace BT;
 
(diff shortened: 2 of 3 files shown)


facontidavide:
problem found and fixed. Thanks for taking the time to report this!

ephson24:
> problem found and fixed. Thanks for taking the time to report this!

One more thing I just realized. The issue persists if you have for example an async node as the last child of the ReactiveSequence. I haven't been able to look into it yet but I think it happens when the child returns RUNNING as I am doing in my **Continue** node.

`<BehaviorTree ID="PowerManagerT">
    <ReactiveSequence>
      <Script code=" LOW_BATT:=20.0; isLowBattery:=false "/>
      <CheckLevel deviceType="BATT"
                  percentage="{LOW_BATT}"
                  result="{isLowBattery}"/>
      <SendWarning deviceType="BATT" deviceStatus="{isLowBattery}" _skipIf="!isLowBattery"/>
      <Continue />
    </ReactiveSequence>
  </BehaviorTree>`

PS: Also run within a while loop

facontidavide:
Trying to understand this... 

1. Does `Continue` finally return SUCCESS, after a while?
2. What is exactly the behavior you get here?

In this logic, EVERY time that `Continue` returns RUNNING, you should execute **Script**, **CheckLevel** and **SendWarning** (if not skipped).

You can not count the number of `tickOnce`, because it is unclear how many times `Continue` will need to be ticked.

Also, keep in mind that **ReactiveSequence** should have **only one** async child

ephson24:
> Trying to understand this...
> 
> 1. Does `Continue` finally return SUCCESS, after a while?
> 2. What is exactly the behavior you get and what do you expect to happen instead?
> 
> In this logic, EVERY time that `Continue` returns RUNNING, you should execute **Script**, **CheckLevel** and **SendWarning** (if not skipped).
> 
> You can not count the number of `tickOnce`, because it is unclear how many times `Continue` will need to be ticked.
> 
> Also, keep in mind that **ReactiveSequence** should have **only one** async child

1. Continue never returns SUCCESS. 
2. The idea is to keep this branch/tree running forever in parallel with the main robot behavior

Yes, Continue is the only Async node in this tree.

facontidavide:
problem found, I think, it is a bug in **SimpleActionNode**.

I will

PR #551, opened 2023-05-01
fix tests with colcon not being run

Rid0n:
I ran into a problem when trying to run `colcon test --packages-select behaviortree_cpp` where the tests wouldn't actually be executed.

Turns out `ament_add_gtest_executable` doesn't go so far as to register the executable as tests preventing colcon from seeing them. `ament_add_gtest` does and the tests now run successfully. 

Read about it here:
[Relevant question on ros answers ](https://answers.ros.org/question/399412/colcon-test-not-running-gtests/)


facontidavide:
thanks a lot

PR #560, opened 2023-05-11
SimpleThreadedActionNode that wraps given functor in ThreadedAction for execution in a separate thread

jmarsik:
Also provides `registerSimpleThreadedAction` method similar to `registerSimpleAction` to quickly register lambdas executing in a separate thread for use in behavior trees.

When implementing this I noticed that `ThreadedAction` name does not conform to the naming convention of other classes in the same "functionality group" ... should I change it to `ThreadedActionNode` and perhaps also update it in the documentation?

### DIFF
--- a/include/behaviortree_cpp/action_node.h
+++ b/include/behaviortree_cpp/action_node.h
@@ -141,6 +141,40 @@ using AsyncActionNode = ThreadedAction;
 #endif
 
 /**
+ * @brief The SimpleThreadedActionNode provides an easy to use ThreadedActionNode.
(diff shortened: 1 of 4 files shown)


jmarsik:
Friendly bump @facontidavide

jmarsik:
Is there anything I can do to merge this PR? @facontidavide

facontidavide:
thanks a lot, but this is not a feature that I want to integrate.

The HARD part of multithreading is how to implement "halt". Without that, all the Reactive behaviors will be broken.

Many user (and this PR) forget that.

PR #570, opened 2023-05-26
Fix Blackboard::set() for StringView

galou:
The second call to `Blackboard::set()` with the same key was not working for `StringView` because `port_info.parseString(value)` returns an empty `Any`.
`Any(std::string(value))` is used instead for `StringView` when the key already exists.

### DIFF
--- a/include/behaviortree_cpp/blackboard.h
+++ b/include/behaviortree_cpp/blackboard.h
@@ -198,6 +198,7 @@ public:
       Any new_value(value);
 
       // special case: entry exists but it is not strongly typed... yet
+      // std::string_view entries are the only entries that are strongly-typed from the beginning.
       if (!port_info.isStronglyTyped())
       {
         // Use the new type to create a new entry that is strongly typed.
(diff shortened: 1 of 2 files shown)


facontidavide:
thanks. Do you have a simple unit test in mind that reproduce the error and fails before this change is applied?

facontidavide:
if you just show me a plain example, I can do it. 
otherwise, add it here: https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/tests/gtest_blackboard.cpp

galou:
Test case added.

facontidavide:
The problem was real (thanks for reporting), but your solution would have introduced a bug.

I fixed it differently here: https://github.com/BehaviorTree/BehaviorTree.CPP/commit/d28e5be842de23235aa688327ebc83c5a3314243

galou:
Thanks for the fix! Just by curiosity and lack of time to study it myself, what was the introduced bug?

PR #571, opened 2023-05-26
Add the AlwaysRunning node

galou:
Use case: throttle a node to run at a given frequency:
```xml
<Repeat num_cycles="1000">

  <ReactiveSequence>

    <Inverter>
      <Timeout msec="1000">
	<AlwaysRunning/>
      </Timeout>
    </Inverter>

    <Log severity="info" message="Throttled message" />

  </ReactiveSequence>

</Repeat>
```


facontidavide:
Actually, there is an easier way, that is using the `TestNode`

```cpp
BT::TestNodeConfig wait_config;
wait_config.async_delay = std::chrono::milliseconds(300);
wait_config.return_status = NodeStatus::SUCCESS;

factory.registerNodeType<BT::TestNode>("Wait", wait_config);
```

facontidavide:
I would rather implement a `Sleep` action, if we want to have a configurable time

PR #575, opened 2023-06-06
Fix missing attribute in generated XML (writeTreeNodesModelXML)

OPerepadia:
Fix missing attribute [BTCPP_format="4"] in the XML generated by the writeTreeNodesModelXML function. This eliminates Groot 2 complaints when importing XML models.


facontidavide:
thanks

PR #579, opened 2023-06-07
changing resetStatus to public

SteveMacenski:
Needed for unit test coverage to reset the root nodes. I can appreciate that this is to prevent users from doing unwise things, but I think at least a reset option is a necessary public method even if `setStatus` is not.

### DIFF
--- a/include/behaviortree_cpp_v3/tree_node.h
+++ b/include/behaviortree_cpp_v3/tree_node.h
@@ -182,6 +182,9 @@ public:
   // Notify the tree should be ticked again()
   void emitStateChanged();
 
+  /// Equivalent to setStatus(NodeStatus::IDLE)
+  void resetStatus();
+
 protected:
   /// Method to be implemented by the user
   virtual BT::NodeStatus tick() = 0;
@@ -222,9 +225,6 @@ private:
   PostTickOverrideCallback post_condition_callback_;
 
   std::shared_ptr<WakeUpSignal> wake_up_;
-
-  /// Equivalent to setStatus(NodeStatus::IDLE)
-  void resetStatus();
 };
 
 //-------------------------------------------------------


SteveMacenski:
@facontidavide let me know if you have thoughts (I'm sure you will :wink:)

facontidavide:
If you want to rest the root node, why not using `Tree::haltTree`.

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/include/behaviortree_cpp/bt_factory.h#L127

The name is not that intuitive, but it will do **exactly** what you want to do.

facontidavide:
I know it is not the subject of this PR, but I find this very confusing

https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/decorator/single_trigger_node.cpp#L31

It is hard for me to see how this has ANY effect in practice. The parent of `SingleTrigger` will almost certainly set it to IDLE almost immediately, making the entire logic pointless.

The only place where it MAY make sense if in a `ReactiveSequence`, but you are basically converting it into a regular `Sequence`.

About the PR, if you make it protected, I will accept it, since it is protected in BT.CPP 4.X

SteveMacenski:
On the PR: done!

On otherwise:  In case you want to do something like "no replanning" so you place the planner BT node under this "only once" so that its skipped for all over traversals of the tree during a particular run while its parent node is `RUNNING`. There are potentially several things like this that for a given task you need to insert a task somewhere in the middle of a BT for its logical use, but it should only do so one time within that particular context. 

I agree it has some limited use due to the IDLE reset of the parent, but for the example below, if we replaced the Rate Controller (orange) with this node, then it would trigger one time only for planning. The IDLE reset actually works in our favor because the parent of that node is always RUNNING except when there's a system level failure exiting the "navigation" side of the tree for the "recovery" side of the tree. Then once we come back, the reset to IDLE makes it so that we retrigger planning 1 time again to continue the task in the new state. 


image

facontidavide:
> so you place the planner BT node under this "only once" so that its skipped for all over traversals of the tree during a particular run while its parent node is RUNNING. 

I am confused... the most basic nodes, `Sequence` or `Fallback`, do that by default.

Now, what I understand is that probably you are trying to deal with a different logic introduced by **your** custom Control Nodes which don't follow the same rules as the "official ones" in BT.CPP.

In your example, you should just replace the `PipelineSequence` (parent of `RateController`) with a regular Sequence and that's it (this is my understanding). In that way you `ComputePathToPose`  once, and never tick that again while `FollowPath` is RUNNING

But probably I am missing something, so... I guess you know what you are doing!

SteveMacenski:
> But probably I am missing something, so... I guess you know what you are doing!

Perhaps, but I also didn't write the PipelineSequence node but I did have a section of the website written up explaining it https://navigation.ros.org/behavior_trees/overview/nav2_specific_nodes.html#control-pipelinesequence

It might be that there was a different / better way to do it, but its been working for so long I haven't had a reason to argue with it. I think the same could be useful with default control nodes, but certainly this is a quirk that is most interesting to us due to it. 

> In your example, you should just replace the PipelineSequence (parent of RateController) with a regular Sequence and that's it (this is my understanding). In that way you ComputePathToPose once, and never tick that again while FollowPath is RUNNING

True. But its still a general primitive that has utility. There were some users asking for it, but I don't recall the exact context at this moment in their applications.

---

@facontidavide can you do a 3.8 release so we can get this into binary format & I can merge the test fix into Nav2?

SteveMacenski:
@facontidavide can you cut a release for 3.8.4 and release to bloom so these changes can be reflected?

PR #599, opened 2023-06-28
Fix error when building static library

OPerepadia:
This commit fixes errors when building the library with option `-DBTCPP_SHARED_LIBS=OFF`


OPerepadia:
Please take a look at the changes. I had to make the function "ErrorNotInit" inline to avoid the issue with multiple definition errors.

PR #625, opened 2023-08-02
Refactoring WhileDoElseNode to allow 2 or 3 children

Aglargil:
**Description:**

This PR addresses the constraint in the WhileDoElse node that mandates the presence of three child nodes. The aim is to improve flexibility and align its behavior with that of the IfThenElse node which allows either two or three child nodes.

**Testing:**

Tests have been conducted to verify the behavior with both two and three child nodes.

Looking forward to your feedback, thank you!

### DIFF
--- a/src/controls/while_do_else_node.cpp
+++ b/src/controls/while_do_else_node.cpp
@@ -29,9 +29,9 @@ NodeStatus WhileDoElseNode::tick()
 {
   const size_t children_count = children_nodes_.size();
 
-  if (children_count != 3)
+  if (children_count != 2 && children_count != 3)
   {
-    throw std::logic_error("WhileDoElse must have 3 children");
+    throw std::logic_error("WhileDoElseNode must have either 2 or 3 children");
   }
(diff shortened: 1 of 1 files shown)


facontidavide:
Side note.

With this change, It seems to me that **WhileDoElse** becomes EXACTLY equal to **ReactiveSequence** with two children.

I am not sure if this may create some confusion. What do you think?

Aglargil:
Thank you for your reply!

I see that **WhileDoElse** is described as

> It is a REACTIVE node of IfThenElseNode

And **IfThenElse** behaves like Sequence when there are only 2 nodes, so it is very design intuitive to make **WhileDoElse** behave consistently with **ReactiveSequence** when there are only two nodes is very intuitive to design, and such duality looks very elegant.

On the flip side, **WhileDoElse** and **IfThenElse** provide better extensibility. For tasks that require a precondition and then execute, we'll likely add the tasks that Else requires to execute later.

PR #658, opened 2023-09-14
Implement starting point for sequence and sequence with memory

5730289021-NN:
With this implementation, one can start at any child node instead of first node only.

### DIFF
--- a/include/behaviortree_cpp/controls/sequence_node.h
+++ b/include/behaviortree_cpp/controls/sequence_node.h
@@ -34,15 +34,18 @@ namespace BT
 class SequenceNode : public ControlNode
 {
 public:
-  SequenceNode(const std::string& name,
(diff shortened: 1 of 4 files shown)


facontidavide:
Look at CI...

Also, I am not changing something as fundamental as the sequence and I am not sure what the motivation is.

5730289021-NN:
It's just a modification that allows you to start at any point, you want for example.

```
<Sequence Start="4">
  <DoA />
  <DoA />
  <DoA />
  <DoA />
  <DoB />
  <DoC />
</Sequence>
```

You can shortcut a process to `<DoB />` without having a lot of `_skipIf` in `<DoA />`

facontidavide:
If you want this functionality, you should create a new Control node, not modify an existing one

PR #669, opened 2023-10-12
feat: improve runtime error output for missing manifest in xml_parser

fmros:
By adding the element ID for which the parser failed to the output it is easier to understand possible XML parsing failures.

### DIFF
--- a/src/xml_parsing.cpp
+++ b/src/xml_parsing.cpp
@@ -600,7 +600,9 @@ TreeNode::Ptr XMLParser::PImpl::createNodeFromXML(const XMLElement* element,
   {
     if(!manifest)
     {
-      throw RuntimeError("Missing manifest. It shouldn't happen. Please report this issue");
+      std::stringstream ss; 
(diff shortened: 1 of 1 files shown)


src/xml_parsing.cpp
```diff
@@ -600,7 +600,9 @@ TreeNode::Ptr XMLParser::PImpl::createNodeFromXML(const XMLElement* element,
   {
     if(!manifest)
     {
-      throw RuntimeError("Missing manifest. It shouldn't happen. Please report this issue");
+      std::stringstream ss; 
+      ss << "Missing manifest for element_ID: " << element_ID << ". It shouldn't happen. Please report this issue.";
+      throw RuntimeError(ss.str());
```
facontidavide, line 605:
```suggestion
      auto msg = StrCat("Missing manifest for element_ID: ", element_ID, ". It shouldn't happen. Please report this issue.");
      throw RuntimeError(msg);
```

facontidavide (changes requested):

PR #675, opened 2023-10-20
Use feature test macro to check availability of `std::from_chars`

chhtz:
Testing for proper availability of `std::from_chars` is best done by testing

    #if __cpp_lib_to_chars >= 201611L

according to https://en.cppreference.com/w/cpp/utility/from_chars.
I don't know what the motivation of checking `__has_include(<GL/gl.h>)` was.

Without this test, using `from_chars` fails to build on older gcc and clang versions:
https://godbolt.org/z/18nPWehaq


facontidavide:
Thanks. Wrong copy and paste

chhtz:
Thanks for the fast merge!

PR #699, opened 2023-11-13
Implement writeTreeDTD()

galou:
Add `<?xml-model href="tree.dtd" type="application/xml-dtd"?>`  or `<!DOCTYPE root SYSTEM "tree.dtd">` to use the benefits of DTD.

### DIFF
--- a/include/behaviortree_cpp/xml_parsing.h
+++ b/include/behaviortree_cpp/xml_parsing.h
@@ -61,6 +61,16 @@ std::string writeTreeNodesModelXML(const BehaviorTreeFactory& factory,
                                    bool include_builtin = false);
 
 /**
+ * @brief writeTreeDTD generates a DTD for the nodes defined in the factory
+ *
+ * @param factory          the factory with the registered types
+ *
+ * @return  string containing the XML.
+ */
+[[nodiscard]]
+std::string writeTreeDTD(const BehaviorTreeFactory& factory);
+
+/**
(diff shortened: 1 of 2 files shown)


galou:
I'm actually not sure whether `<include/>` is allowed within `<TreeNodesModel/>` or `<BehaviorTree/>`.

galou:
What about `<SubTree/>` within `<TreeNodesModel/>`? If they are accepted, I have the issue that DTD is not powerful enough to describe differently a `<SubTree/>` within `<root/>` and within `<TreeNodesModel/>` (I don't know much about DTD).

galou:
@facontidavide what is blocking merging?

facontidavide:
Could you build an example showing how this works in the folder [examples](https://github.com/BehaviorTree/BehaviorTree.CPP/tree/master/examples)?
Something like **ex06_something_something.cpp**

> I'm actually not sure whether <include/> is allowed within <TreeNodesModel/> or <BehaviorTree/>

It is not

> What about <SubTree/> within <TreeNodesModel/>? If they are accepted, I have the issue that DTD is not powerful enough to describe differently a <SubTree/> within <root/> and within <TreeNodesModel/> (I don't know much about DTD).

It is now. Look at example [ex05_subtree_model](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/examples/ex05_subtree_model.cpp)

PR #708, opened 2023-12-05
Fix any istype

galou:


facontidavide:
Motivation?

galou:
I need to convert blackboard entries to my Lua environment. I cannot use `tryCast` because a lot of things cast to a lot of things. `isType<bool>()` and similar returned `false` because of the type casting  inside `Any` (cf. unit-test). The second commit fixes this.

include/behaviortree_cpp/utils/safe_any.hpp
```diff
@@ -269,15 +276,15 @@ inline Any &Any::operator =(const Any &other)
 
 inline bool Any::isNumber() const
 {
-  return _any.type() == typeid(int64_t) ||
-         _any.type() == typeid(uint64_t) ||
-         _any.type() == typeid(double);
+  return type() == typeid(int64_t) ||
+         type() == typeid(uint64_t) ||
+         type() == typeid(double);
```
facontidavide, line 281: Not sure about this. Shouldn't you include all the possible arithmetic types?

facontidavide (changes requested):

include/behaviortree_cpp/utils/safe_any.hpp
galou: Or use `_any.type()`?

Now the `isNumber` function is not called in behaviortree_cpp (`isCastedAsNumber` is used instead).

facontidavide:
this is still public API, I want to be sure that the behavior of public method is not changed

facontidavide:
I don't want to introduce the method isCastedType, therefore I cherry picked the other changes.

Thanks!

Issue #706, opened 2023-11-30
Enable local variables in sub-tree


galou:
As sub-trees get more complicated, I tend to always use `_autoremap="true"` rather than writing a long list of remappings. Even if I would be OK with not using `_autoremap`, I cannot ensure that other users of the sub-tree I write also avoid to use it.

Now, I have a case where I compute some small trajectories identified with a shared prefix and then combine them. I do not know in advance how many of these trajectories will be in the blackboard but what I know for sure is that I don't want to share these small trajectories with other calls to the sub-tree.

I see two solutions:
- use a prefix to identify a local variable, e.g. `local{var}`
- add an attribute to the `BehaviorTree` element to list the variables that should not be shared even if `_autoremap` is used. Not so nice, IMHO.

I can implement something but I'd like to have some guidance on how to implement this.

facontidavide:
this seems like a very specific use case. I understand  that in your case could be annoying, but it is hard for me to find an elegant solution.

The prefix is doable, but seems too specific.

What about excluding from auto-remapping all the port names that start with "_",

"I_am_public"
"_I_am_private"

This feels very familiar for people developing C++

facontidavide:
What do you think?

galou:
This would be a useful addition that would help avoiding potential issues when using subtrees.

facontidavide:
done

galou:
Woah, that was quick! Thanks!

facontidavide:
That's what my wife said, excluding "thanks"

PR #709, opened 2023-12-05
Add the possibility to remove a blackboard entry.

galou:
Add a node to use this functionality.


galou:
This is related to #706 . I need the ability to have "local" variables. Instead of variable scope, I can now delete some blackboard entries at the end of my subtree so that I have a clean blackboard in the case that the subtree is called again.

facontidavide:
this looks useful. 
See my suggested change in the copyright. After that fix, I will merge it

facontidavide (changes requested):

facontidavide:
I was able to commit that myself. Merging as soon as Ci turns green

corot:
Agree this looks useful. But wouldn't be neater to use SetBackboard with an empty input?
Would be that possible?
:thinking:

facontidavide:
merged with some changes. Thanks

PR #730, opened 2023-12-21
Add optional metadata to TreeNodeManifest

adlarkin:
Motivation: Users may want to specify additional data for a node ("metadata"). They now have the option to do this, and this "metadata" will be a part of the automatically created XML when calling [writeTreeNodesModelXML](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/ceb25a7f78bc88b347d903b78969e2bd68f7df4b/src/xml_parsing.cpp#L1093).

### DIFF
--- a/include/behaviortree_cpp/basic_types.h
+++ b/include/behaviortree_cpp/basic_types.h
@@ -10,7 +10,9 @@
 #include <chrono>
 #include <memory>
 #include <string_view>
+#include <utility>
 #include <variant>
+#include <vector>
 #include <optional>
 
 #include "behaviortree_cpp/utils/safe_any.hpp"
@@ -450,6 +452,43 @@ struct has_static_method_description<
 {
 };
 
+/// Optional metadata for a TreeNodeManifest.
+/// The metadata can represent an xml element with text:
+///   <metadata_name>Descriptive text</metadata_name>
+/// Or an xml element with an attribute:
+///   <metadata_name attribute="value"/>
+struct ManifestMetadata
+{
+  /// A pair containing the <name, value> of a metadata's XML attribute.
+  using Attribute = std::pair<std::string, std::string>;
+
+  [[nodiscard]] bool representsText() const
+  {
+    return std::holds_alternative<std::string>(text_or_attribute);
+  }
+
+  [[nodiscard]] bool representsAttribute() const
+  {
+    return std::holds_alternative<Attribute>(text_or_attribute);
+  }
+
+  std::string name;
+  std::variant<std::string, Attribute> text_or_attribute;
+};
+
+template <typename T, typename = void>
+struct has_static_method_metadata : std::false_type
+{
+};
+
+template <typename T>
+struct has_static_method_metadata<
+    T, typename std::enable_if<
+           std::is_same<decltype(T::metadata()), std::vector<ManifestMetadata>>::value>::type>
+  : std::true_type
+{
+};
+
 template <typename T> [[nodiscard]]
 inline PortsList getProvidedPorts(enable_if<has_static_method_providedPorts<T>> = nullptr)
 {
@@ -467,4 +506,3 @@ using TimePoint = std::chrono::high_resolution_clock::time_point;
 using Duration = std::chrono::high_resolution_clock::duration;
 
 }   // namespace BT
-
--- a/include/behaviortree_cpp/bt_factory.h
+++ b/include/behaviortree_cpp/bt_factory.h
@@ -44,13 +44,20 @@ template <typename T>
 inline TreeNodeManifest CreateManifest(const std::string& ID,
                                        PortsList portlist = getProvidedPorts<T>())
 {
-  if constexpr( has_static_method_description<T>::value)
+  if constexpr( has_static_method_description<T>::value && has_static_method_metadata<T>::value )
   {
-    return {getType<T>(), ID, portlist, T::description()};
+    return {getType<T>(), ID, portlist, T::description(), T::metadata()};
   }
-  else {
-    return {getType<T>(), ID, portlist, {}};
+  else if constexpr( has_static_method_description<T>::value && !has_static_method_metadata<T>::value )
+  {
+    return {getType<T>(), ID, portlist, T::description(), {}};
   }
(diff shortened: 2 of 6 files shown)


include/behaviortree_cpp/basic_types.h
```diff
@@ -450,6 +452,43 @@ struct has_static_method_description<
 {
 };
 
+/// Optional metadata for a TreeNodeManifest.
+/// The metadata can represent an xml element with text:
+///   <metadata_name>Descriptive text</metadata_name>
+/// Or an xml element with an attribute:
+///   <metadata_name attribute="value"/>
```
sea-bass, line 459: Do we also want to support the following case?

```
 <metadata_name attribute1="value1" attribute2="value2"/>
```

If so, might be worth the `std::variant` in question to use `std::vector<Attribute>` rather than (or in addition to) `Attribute` and it all may just shake out.

include/behaviortree_cpp/bt_factory.h
```diff
@@ -44,13 +44,20 @@ template <typename T>
 inline TreeNodeManifest CreateManifest(const std::string& ID,
                                        PortsList portlist = getProvidedPorts<T>())
 {
-  if constexpr( has_static_method_description<T>::value)
+  if constexpr( has_static_method_description<T>::value && has_static_method_metadata<T>::value )
```
sea-bass, line 47: I think from our last discussion with @facontidavide we said we'd just replace description with metadata instead of adding it, right?

Where the description would just be one element of the metadata, if users wanted it?

include/behaviortree_cpp/tree_node.h
```diff
@@ -39,6 +40,7 @@ struct TreeNodeManifest
   std::string registration_ID;
   PortsList ports;
   std::string description;
+  std::vector<ManifestMetadata> metadata;
```
facontidavide, line 43: As mentioned earlier, I prefer simplicity and to keep it:
```cpp
std::vector<std::pair<std::string, std::string>> metadata;
//or
struct Metadata {
  std::string name;
  std::string value;
};

std::vector<Metadata> metadata;
```

include/behaviortree_cpp/bt_factory.h
facontidavide: yes

src/xml_parsing.cpp
```diff
@@ -989,6 +989,24 @@ void addNodeModelToXML(const TreeNodeManifest& model,
     element->InsertEndChild(description_element);
   }
 
+  for (const auto& metadata : model.metadata)
```
facontidavide, line 992: The generated XML should be

```XML
<MetadataFields>
  <Metadata name="foo" value="hello"/>
  <Metadata name="bar" value="42"/>
</MetadataFields>

File: include/behaviortree_cpp/basic_types.h [hunk 5]
```diff
@@ -450,6 +452,43 @@ struct has_static_method_description<
...
+  }
+
+  [[nodiscard]] bool representsAttribute() const
+  {
+    return std::holds_alternative<Attribute>(text_or_attribute);
+  }
+
+  std::string name;
+  std::variant<std::string, Attribute> text_or_attribute;
+};
```
facontidavide, line 477: Too much complexity for a niche feature, in my opinion.
I would keep it simple and support only the attribute case.
Therefore:
```cpp
struct Metadata {
  std::string name;
  std::string value;
};
```

facontidavide (changes requested):
Added some request for changes.

include/behaviortree_cpp/basic_types.h
adlarkin: Sounds like we don't need the variant anymore once I implement the approach Davide asked for in his review.

include/behaviortree_cpp/tree_node.h
adlarkin: Sounds good! Made this change in https://github.com/BehaviorTree/BehaviorTree.CPP/pull/730/commits/96910f265786458ac2f1b33a9f3d1a3adaf26473

include/behaviortree_cpp/bt_factory.h
adlarkin: I replaced `description` with `metadata` in https://github.com/BehaviorTree/BehaviorTree.CPP/pull/730/commits/96910f265786458ac2f1b33a9f3d1a3adaf26473

src/xml_parsing.cpp
adlarkin: Done! https://github.com/BehaviorTree/BehaviorTree.CPP/pull/730/commits/96910f265786458ac2f1b33a9f3d1a3adaf26473

include/behaviortree_cpp/bt_factory.h
```diff
@@ -44,13 +46,11 @@ template <typename T>
...
-  if constexpr( has_static_method_description<T>::value)
+  if constexpr( has_static_method_metadata<T>::value )
   {
-    return {getType<T>(), ID, portlist, T::description()};
-  }
-  else {
-    return {getType<T>(), ID, portlist, {}};
+    return {getType<T>(), ID, portlist, T::metadata()};
   }
+  return {getType<T>(), ID, portlist, {}};
```
sea-bass, line 53: not that it matters, but the previous version of the code explicitly used if-else instead of relying on the `if` condition to return. Of course, it's up to what the maintainer prefers :)

src/xml_parsing.cpp
```diff
@@ -982,11 +982,18 @@ void addNodeModelToXML(const TreeNodeManifest& model,
...
 
-  if (!model.description.empty())
+  if (!model.metadata.empty())
   {
-    auto description_element = doc.NewElement("description");
-    description_element->SetText(model.description.c_str());
-    element->InsertEndChild(description_element);
+    auto metadata_root = doc.NewElement("MetadataFields");
+
+    for (const auto& [name, value] : model.metadata)
```
sea-bass, line 989: this is a beautiful line of C++

sea-bass (commented):
LGTM! Though of course approval is not on me here.

Issue #710, opened 2023-12-05
Enums evaluate to double not integral?


Roboterbastler:
I have come across the problem that I initialized a blackboard entry from an enum value (registered with `registerScriptingEnums`) but the entry seems to have real number type afterwards and a SwitchX node doesn't work properly because of this (string equality is not given, e.g. "0.0000" != "0").

I tried to trace this down a bit to this line and am wondering why it is casted to double and not an integral type, is there a reason?

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/ae95e680ab3c58f1e773f05d6372c10f41b546e8/include/behaviortree_cpp/scripting/operators.hpp#L75

facontidavide:
This is a known limitation of the scripting language, sorry. I will try to find a solution.
It would be nice if you can prepare failing unit test!!! :pray:

Roboterbastler:
Sorry for the late response, I proposed a failing unit test in #745.

facontidavide:
Thanks. I will try to address this in my next release cycle

facontidavide:
fixed. Thanks for the help. See unit test added for correct usage

PR #745, opened 2024-01-16
Add unit test for switch node in combination with enum values

Roboterbastler:
This PR adds a failing unit test for using an enum with the Switch node (see issue #710).

This test succeeds if you change line 98 in the test to

`<Script code=" my_color := 1 "/>`

and line 101 to

`case_2="1"`

### DIFF
--- a/tests/gtest_enums.cpp
+++ b/tests/gtest_enums.cpp
@@ -89,3 +89,34 @@ TEST(Enums, StrintToEnum)
   }
 }
 
+TEST(Enums, SwitchNodeWithEnum)
+{
+  const std::string xml_txt = R"(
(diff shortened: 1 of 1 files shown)


facontidavide:
This syntax is invalid. It should be:

```cpp
<Switch4 variable="{my_color}"
  case_1="Red"
  case_2="Blue"
  case_3="Green"
   case_4="Undefined">
         <AlwaysFailure name="case_red" />
         <AlwaysSuccess name="case_blue" />
         <AlwaysFailure name="case_green" />
         <AlwaysFailure name="case_undefined" />
         <AlwaysFailure name="default_case" />
 </Switch4>
```

PR #746, opened 2024-01-17
Use snprintf instead of sprintf

galou:
- Augment the buffer size on doc error.
- Let sprintf in switch_node.h since the max. string length is known.


facontidavide:
thanks

Issue #758, opened 2024-01-31
change signature of registerNodeType to accept StringView


stewpend0us:
As far as I can tell...right now ```registerNodeType``` requires a ```const std::string& ID``` but ultimately just makes a copy of the string anyway. Why not use StringView instead?

That way I can keep a ```static constexpr StringView ID``` on my class and pass it directly into ```registerNodeType<MyType>(MyType::ID, ...)``` without having to jump through the (very small) extra hoop of defining a ```static const std::string```

Better yet (maybe?) allow me to optionally define my class with a ```static constexpr StringView ID``` property that _just_ gets picked up by the register function?

facontidavide:
feel free to submit a PR

facontidavide:
So, I see that there **could** be 1 copy of a string less during initialization, but since this run-time optimization only happens at the very beginning, when the node is registered, it is not a good reason to apply a non-ABI compatible change

stewpend0us:
Sorry. I can do what I need with the "registrationName()" method.

PR #759, opened 2024-02-01
switch some low-hanging fruit from 'const std::string&' to 'const StringView'

stewpend0us:
const std::string& runs deep...hard to change a whole lot without making breaking changes to or extending the API. This should 'fix' #758 which is all I was really after.

Is updating the signature of everyone's constructors to take StringView instead of std::string& worth working on?

### DIFF
--- a/include/behaviortree_cpp/bt_factory.h
+++ b/include/behaviortree_cpp/bt_factory.h
@@ -42,14 +42,14 @@ inline NodeBuilder CreateBuilder(Args... args)
 }
 
 template <typename T>
-inline TreeNodeManifest CreateManifest(const std::string& ID,
+inline TreeNodeManifest CreateManifest(const StringView ID,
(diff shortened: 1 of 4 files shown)


facontidavide:
After reviewing your change, I don't see how this will make the user's life easier

stewpend0us:
The benefit is very minor.

Right now I can't do this:
```C++
class MyClass {
public:
    static const std::string ID = "MyClassXMLID"; // compiler error (have to put the definition somewhere else)
    static constexpr std::string_view ID = "MyClassXMLID"; // this is fine
...
}

...

factory_.registerNodeType<MyClass>(MyClass::ID); // only works when ID is a string not a string_view
```

Is it minor and kind of dumb? yes. But, as a user, I'd prefer to be able to do the in-line constexpr definition of ID.

Issue #766, opened 2024-02-13
Allow concatenation of string and numbers in scripting


bochen87:
Currently, only concatenation of strings is allowed. Trying to concatenate a number and a  string leads to a runtime error https://github.com/BehaviorTree/BehaviorTree.CPP/blob/58326a8a08d9c51e2ed8674aef3710bb6209867b/include/behaviortree_cpp/scripting/operators.hpp#L255

It would be nice to also allow concatenation of string and numbers

facontidavide:
I can see the use case, I will put this on my list of "todo".

bochen87:
> I can see the use case, I will put this on my list of "todo".

I made a PR https://github.com/BehaviorTree/BehaviorTree.CPP/pull/771

PR #771, opened 2024-02-13
Add support to concat strings and numbers in scripting

bochen87:
Solves #766

### DIFF
--- a/include/behaviortree_cpp/scripting/operators.hpp
+++ b/include/behaviortree_cpp/scripting/operators.hpp
@@ -252,7 +252,9 @@ struct ExprBinaryArithmetic : ExprBase
         }
       }
     }
-    else if (rhs_v.isString() && lhs_v.isString() && op == plus)
+    else if (op == plus && ((rhs_v.isString() && lhs_v.isString()) ||
+                            (rhs_v.isString() && lhs_v.isNumber()) ||
+                            (rhs_v.isNumber() && lhs_v.isString())))
     {
       return Any(lhs_v.cast<std::string>() + rhs_v.cast<std::string>());
     }


facontidavide:
The potential issue I see with this is that there are two contraddicting use case:

1. Someone may want to concatenate string with numbers, so for instance

```
"path_" + 4 -> "path_4"
```

2. Someone may want strings to be interpreted as numbers
```
"3" + 4 -> 7
```

What would be probably "wrong" is:
```
"3" + 4 -> "37"
```
So, the only sane way to do this, in my opinion is including functions `to_number()` and `to_string()`:

```
"path" + to_string(4) -> "path_4"
to_number("3") + 4 -> 37
```
This is considerably more work, but the only way to avoid technical debt from being created, in my opinion.

bochen87:
> The potential issue I see with this is that there are two contraddicting use case:
> 
> 1. Someone may want to concatenate string with numbers, so for instance
> 
> ```
> "path_" + 4 -> "path_4"
> ```
> 
> 2. Someone may want strings to be interpreted as numbers
> 
> ```
> "3" + 4 -> 7
> ```
> 
> What would be probably "wrong" is:
> 
> ```
> "3" + 4 -> "37"
> ```
> 
> So, the only sane way to do this, in my opinion is including functions `to_number()` and `to_string()`:
> 
> ```
> "path" + to_string(4) -> "path_4"
> to_number("3") + 4 -> 37
> ```
> 
> This is considerably more work, but the only way to avoid technical debt from being created, in my opinion.

It would be nice, but I guess quite a large amount of work to make it work properly, as this approach also has some pitfalls, i.e. "3,14" vs "3.14" vs "8'000". However, in the meantime, this could still be a good enough intermediate solution as it does at least allow for the usecase 1 to be solved.

facontidavide:
> However, in the meantime, this could still be a good enough intermediate solution as it does at least allow for the usecase 1 to be solved.

Sorry, not adding this feature yet. I can't afford to have side effects that haven't been considered carefully

seanng-1:
Would a better solution for this be: to add a new operator specifically for string concatenation? Similar to how it is done in Lua.
e.g.
```
value := 10
message := "The value is " .. value
```
would result in `The value is 10`

I can see this avoiding any issues with people wanting to use strings converted to numbers as it is explicitly for string concatenation.

facontidavide:
closing in favor of https://github.com/BehaviorTree/BehaviorTree.CPP/pull/802

PR #772, opened 2024-02-13
Initialize all instances of `std::atomic_bool`

MarqRazz:
I have been testing a sequence of behaviors based on the `StatefulActionNode` and noticed that sometimes my behaviors would immediately return because the value of `isHaltRequested()` was checked in the `onRunning()` method. After some digging I found that the value  `halt_requested_` is uninitialized and set to a random value when constructing my `StatefulActionNode` based objects.

I added a simple print statement to the constructor of `StatefulActionNode` and this is what I found:
```
[node_main] StatefulActionNode: halt_requested_ = 0
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] StatefulActionNode: halt_requested_ = 0
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] StatefulActionNode: halt_requested_ = 0
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] StatefulActionNode: halt_requested_ = 1
[node_main] : Sequence IDLE -> RUNNING
....
```

This PR initializes all instances of `std::atomic_bool` and a handful of `bool` member variables. I would think it's best practice to initialize all member variables but wanted to push this up and get your thoughts before investing more time.

### DIFF
--- a/examples/broken_sequence.cpp
+++ b/examples/broken_sequence.cpp
@@ -35,7 +35,7 @@ public:
 
 private:
   int time_;
-  std::atomic_bool stop_loop_;
+  std::atomic_bool stop_loop_ = false;
 };
 
 int main()
--- a/include/behaviortree_cpp/action_node.h
+++ b/include/behaviortree_cpp/action_node.h
@@ -132,7 +132,7 @@ public:
 
 private:
   std::exception_ptr exptr_;
-  std::atomic_bool halt_requested_;
+  std::atomic_bool halt_requested_ = false;
   std::future<void> thread_handle_;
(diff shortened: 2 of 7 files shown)


facontidavide:
wow, thanks for finding (and fixing) this

Issue #768, opened 2024-02-13
Tree Execution Error When Setting `InputPort` Value to Empty String with `std::optional<T>` Default Value as `std::nullopt` After Commit 789ce6ea0ad3627923bd2389b8fb9199ffab6d84


KentaKato:
Following the changes made in commit 789ce6ea0ad3627923bd2389b8fb9199ffab6d84, an error occurs during tree execution when `std::nullopt` is set as the default value for an `InputPort` of type `std::optional<T>`, and the `InputPort`'s value is subsequently set to an empty string ("").

This issue is exemplified when `T` is set to `std::optional<geometry_msgs::Pose>`, resulting in the following error message:

```
You (maybe indirectly) called BT::convertFromString() for type [std::optional<geometry_msgs::Pose_<std::allocator<void> > >], but I can't find the template specialization.

[ERROR] [1707813348.400082594]: [port_utils::getInput] Failed to get input 'poses': You didn't implement the template specialization of convertFromString for this type: std::optional<geometry_msgs::Pose_<std::allocator<void> > >
[ERROR] [1707813348.400098635]: [BT][InputTest][InputTest](UID 1): failed to get inputs
```

In Groot2, when an InputPort doesn't explicitly have a value assigned by the user, the system automatically assigns an empty string (""). However, this automatic assignment leads to runtime errors when the InputPort is of type std::optional<T> with std::nullopt set as its default value. Ideally, the system should detect when an empty string is auto-assigned to such InputPorts and instead utilize the user-defined std::nullopt default value to prevent these runtime errors.

facontidavide:
I Will look into this

facontidavide:
Check 83fce13a38efe81f2dd5713328f94942f32f8c9a

If the problem persists, please tell me how to extend the unit test **Default_Issues_767_768** to cover your case.

KentaKato:
@facontidavide 
Thank you for your prompt response. However, the issue still persists. I have implemented a unit test to reproduce it and submitted a pull request https://github.com/BehaviorTree/BehaviorTree.CPP/pull/773.

PR #773, opened 2024-02-14
Issue #768: Add unit test to reproduce

KentaKato:
Unit test to reproduce issue https://github.com/BehaviorTree/BehaviorTree.CPP/issues/768

### DIFF
--- a/tests/gtest_ports.cpp
+++ b/tests/gtest_ports.cpp
@@ -533,8 +533,66 @@ TEST(PortTest, DefaultInputStrings)
   ASSERT_EQ(status, NodeStatus::SUCCESS);
 }
 
+struct TestStruct
+{
+  int a;
+  double b;
+  std::string c;
+};
 
-TEST(PortTest, Default_Issues_767_768)
+class NodeWithDefaultNullptr : public SyncActionNode
+{
+public:
+  NodeWithDefaultNullptr(const std::string& name, const NodeConfig& config) :
+      SyncActionNode(name, config) {}
+
+  NodeStatus tick() override
+  {
+    std::shared_ptr<TestStruct> test_struct_ptr;
+    if (!getInput("input", test_struct_ptr))
+    {
+      throw std::runtime_error("NodeWithDefaultNullptr: failed input");
+    }
+    if (test_struct_ptr == nullptr)
+    {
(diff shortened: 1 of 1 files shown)


facontidavide:
There are two errors in this test:

1. First in line 577, you should use `std::optional`, not `std::shared_ptr`
2. The following XML is wrong:

```xml
<NodeWithDefaultNullptr input=""/>
```

And should be instead:
```xml
<NodeWithDefaultNullptr />
```
In fact, the former version will override the default value and assigns to the port an **empty string**

KentaKato:
@facontidavide 
Regarding point 2, I understand that to use the default value, the port should not be listed in the XML.

Yet, with Groot2, an empty string is automatically applied to any empty Port fields. Therefore, to ensure that default values are utilized for certain ports when generating a tree with Groot2, would it be accurate to say that the necessary steps are:

1. Outputting XML from Groot2
2. Removing the ports where default values are intended to be used

Is this understanding correct?

facontidavide:
> Yet, with Groot2, an empty string is automatically applied to any empty Port fields

That is an entirely separate problem that I will be happy to fix. 
But I just check and that doesn' happen in Groot2 1.5.2

image

If the default value is specified in the **<TreeNodesModel>** than Groot2 does to add ports with empty strings.

Have a look at commit 6c9929c99e0546fcb51d9a0e2dc1febe5c7a624b , where I cherry-picked some of your code.

You will now be alerted that the default is overridden when the tree is constructed.

KentaKato:
I deeply appreciate the verification and update.

I have understood the specifications of Groot2 version 1.5.2 and have also confirmed that alerts are generated.
Upon further verification, I identified the cause of the issue I was facing.

When generating XML, I have been outputting the TreeNodesModel to the console using BT::writeTreeNodesModelXML and then copying & pasting it into an XML file. It seems that BT::writeTreeNodesModelXML outputs default="" when the default value is not of a primitive type (such as nullptr or std::optional). As a result, it seems that an empty string "" was being assigned as the default value for empty Ports on Groot2.

In cases where a non-primitive type is set as the default value, what procedure is expected for generating the TreeNodesModel?

facontidavide:
Oh, thanks for finding this.

I will have a look and see if I can fix it.

PR #784, opened 2024-03-14
Add unit test for blackboard backup

vicmassy:
This unit test provides an example of how the backup/restore blackboard process creates new variables in the Main Tree


facontidavide:
I cherry-picked your code and applied it here with some changes: https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/tests/gtest_blackboard.cpp#L510-L559

The tests are now passing

PR #789, opened 2024-03-19
Extend unit test for blackboard backup to run the second tree

vicmassy:
Extended the Unit testing of the Blackboard Backup to run the second iteration of the tree after the blackboard restore. Some blackboard variables are not being remapped correctly to the Subtree


facontidavide:
unit test fixed with 55a7b70b7ab6c1f2be811009e67f6e6630839947

PR #792, opened 2024-03-22
Add option to use third party as external dependencies

uilianries:
Greetings!

After the PR https://github.com/conan-io/conan-center-index/pull/20010, I checked a little bit the current state of BehaviorTree, and there is a window to use external dependencies, instead of those vendorized versions. The only exception is for cppzmq, because it has an internal patch. But why moving to external dependencies, if vendorized is working and keeps versions aligned? Well, people will be able to use newer versions, or even use customized versions, without touching this project.

To summarize this PR:

- Add the CMake option BTCPP_VENDOR_3RDPARTY. By default, it will be ON, and will follow the very same behavior as in master branch right now: Consumed all dependencies from 3rdparty folder. When using as OFF, it will use external instead, including:
  - foonathan-lexy/2022.12.1
  - zeromq/4.3.5
  - sqlite3/3.45.2
  - minicoro/0.1.3
  - minitrace/cci.20230905
  - wildcards/1.4.0
  - tinyxml2/10.0.0
  - gtest/1.14.0
  - cppzmq will be from 3rdparty folder always! 

For now, I only checked it with Conan, but could be agnostic, by adding a `find_package()` in CMakeLists.txt and changing the `target_link_libraries`, but would requires more changes, harder to review, but could be a next PR. 

- Change the import headers to not use namespace for some dependencies. Well, sounds contradictory, but real thing is, all those projects install their headers in `include` folder directly, no namespace folder, even their tests and examples are using on that way. Here is some information:

  - minicoro: https://github.com/edubart/minicoro?tab=readme-ov-file#minimal-example
  - wildcards: https://github.com/zemasoft/wildcards?tab=readme-ov-file#basic-usage
  - minitrace: https://github.com/hrydgard/minitrace/blob/master/CMakeLists.txt#L37
  - tinyxml2: https://github.com/leethomason/tinyxml2/blob/master/CMakeLists.txt#L81

- I tested on my machine™, but as it's irrelevant to the future, I added a new CI job to only 
build the project (including unit tests) using only external dependencies. Plus, I updated to use Conan 2.x (Conan 1.x is to be deprecated in the future.)

### DIFF
--- a/.github/workflows/cmake_ubuntu.yml
+++ b/.github/workflows/cmake_ubuntu.yml
@@ -24,13 +24,10 @@ jobs:
       id: conan
       uses: turtlebrowser/get-conan@main
       with:
-        version: 1.59.0
+        version: 2.2.1
 
     - name: Create default profile
-      run: conan profile new default --detect
-
-    - name: Update profile
-      run: conan profile update settings.compiler.libcxx=libstdc++11 default
+      run: conan profile detect
 
     - name: Create Build Environment
       # Some projects don't allow in-source building, so create a separate build directory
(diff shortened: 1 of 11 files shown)


uilianries:
@facontidavide Sorry, still need some touch. I kept in draft to avoid merging, but I'll close for now to avoid triggering the CI and send notification, so I can test in my fork only. I'll re-open after fixing everything.

PR #802, opened 2024-04-07
Add string concatenation operator to scripting

seanng-1:
Helps address #766 but just supports string concatenation resulting in a string.

Adds support for string concatenation via a new operator, similar to how it is done in Lua:

e.g.
```
value := 10
message := "The value is " .. value
```
would result in `The value is 10`

Haven't gone much further than testing for my own use case. I can make some other changes if required so please let me know :)


facontidavide:
Also, I can see Pixi windows failing. Can you check that out, please?

seanng-1:
CI should be passing now :)

PR #804, opened 2024-04-12
Export plugins to share directory & register CrossDoor plugin

MarqRazz:
This PR is proposing to export the plugins to the `package_name/share` directory to make it easy for other ROS packages to find the plugins. It also adds the registration function to the CrossDoor example.


facontidavide:
looks goot to me. waiting for CI to become green

sample_nodes/crossdoor_nodes.cpp
```diff
@@ -74,3 +74,11 @@ void CrossDoor::reset()
   _door_locked = true;
   _pick_attempts = 0;
 }
+
+// This function must be implemented in the .cpp file to create
+// a plugin that can be loaded at run-time
+BT_REGISTER_NODES(factory)
+{
+  static CrossDoor cross_door;
```
MarqRazz, line 82: Can you think of a better pattern to show here that keeps this object in scope (same thing [here](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/3cc7b4e120b20ee54c850ceba564cdd4f17827ea/sample_nodes/dummy_nodes.h#L124))? I plan to make lots of Behaviors that share objects (to keep them from being duplicated across multiple behaviors in the tree) and feel like there might be a better way than making them `static`.

facontidavide: actually, I don't believe this is a good pattern. You should probably create the object "outside" and pass it as an argument.

Let me think about it and propose a different pattern

MarqRazz: To continue with my general executor I'm proposing [here](https://github.com/BehaviorTree/BehaviorTree.CPP/discussions/806) I think we would need the option to transfer the ownership to the factory, or something else within BT.CPP, to keep if from requiring prior knowledge of objects the Behaviors need while loading plugins.

I'm not sure how similar this situation is but one pattern that comes to mind is in MTC. When assembling tasks you create an [object](https://github.com/ros-planning/moveit_task_constructor/blob/master/demo/src/pick_place_task.cpp#L154-L156) to store stages and then [add/move](https://github.com/ros-planning/moveit_task_constructor/blob/master/demo/src/pick_place_task.cpp#L209) unique pointers the user creates to it.

msis:
@facontidavide Can this please be merged now?

axwu:
@facontidavide  can we merge this patch? It's impossible to run BehaviorTree.ROS2 without it, thanks!

Issue #823, opened 2024-05-19
Impossible to write to global blackboard using output ports


robin-mueller:
In my application I am trying to write to a global blackboard entry using `setOutput` of a tree node, but an entry in the local blackboard is created instead.

Let an example be:

```xml
<BehaviorTree ID="MainTree">
    <Sequence>
        <Script code="@global_var:=0"/>
        <SubTree ID="BugExample" />
    </Sequence>
</BehaviorTree>

<BehaviorTree ID="BugExample">
    <Sequence>
        <SetState in="1"/>
        <GetState out="{@global_var}"/>
    </Sequence>
</BehaviorTree>
```

The nodes `SetState` and `GetState` set respectively get the value of an internal state variable using the corresponding ports. With the latest version to date, `global_var` is created in the blackboard of *MainTree* with value 0 and `@global_var` in the blackboard of *BugExample* with value 1. The expected behavior would be that `GetState` updates the entry `global_var` in *MainTree* as I understand it.

I'd suggest to implement an example like this in the associated [test](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/287eb59cefdc375cfd45bfa1431a840163e8199a/tests/gtest_blackboard.cpp#L566-L608) to verify this issue, since I don't see that it is covered yet.

robin-mueller:
I've added the test in the pull request above.

I currently don't have too much time to look into it, but I assume the issue to have its origin [here](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/287eb59cefdc375cfd45bfa1431a840163e8199a/include/behaviortree_cpp/blackboard.h#L203). `storage_` is searched for a key `@global_var` which obviously cannot be found.

facontidavide:
fixed

PR #824, opened 2024-05-19
Test for setting a global blackboard entry using a node's output port

robin-mueller:
This pull request adds the test mentioned in issue #823. The workflow runs show that the test fails. Global blackboard idiom is not working for output ports.


facontidavide:
fixed. Thanks for the help

PR #778, opened 2024-02-23
use ament export targets

Ryanf55:
# Purpose

Upgrade the support to latest ament recommendations, which uses CMake targets.

# Addtions

I added a CI task to check it works. The stuff in `export` folder is merely to check the installation of BTCPP is correct.

# Ticket

Closes #751 

# CI status

CI will be red until the next `humble` sync.

PR #826, opened 2024-05-21
Add in call to ament_export_targets.

clalancette:
That way downstream ament packages can use this
package as a CMake target.

I'll note that I targeted this PR to the v3.8 branch, as that is what I'm compiling against for Navigation2 on Iron right now.  But I'm happy to target to another branch, depending on best development practices for this repository.


Ryanf55:
This is an alternative to my proposal at https://github.com/BehaviorTree/BehaviorTree.CPP/pull/778

facontidavide:
I will add this to the main branch too. thanks

PR #833, opened 2024-06-06
Building with a recent compiler fails due incompatible expected library

psyinf:
When build with a recent compiler (gcc 14.1/clang 17.0.1) the build failed due to the dependency [expected-lite](https://github.com/martinmoene/expected-lite) doesn't provide a make_unexpected when std::expected is included. 
I've contacted the author and proposed a fix. He fixed the library and released v0.8.0.

I've taken the freedom to add the new expected-header to the source tree and also adapted a possible oversight (which also doesn't compile with a newer compiler. If the use of ´using ExpectedEntry = nonstd::expected_lite::expected<Entry, std::string>;´ is intended this needs to be fixed to be compatible within the expected-library, i guess.

Additionally I've added some changes to the CMakeLists file to disable building of the sample nodes if no test, nor examples are needed.

A general question: Currently the dependencies are hard-wired, any opinion towards using [CPM](https://github.com/cpm-cmake/CPM.cmake) or plain cmake FetchContent here?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,7 +194,9 @@ message( STATUS "BTCPP_LIB_DESTINATION:   ${BTCPP_LIB_DESTINATION} " )
 message( STATUS "BTCPP_INCLUDE_DESTINATION: ${BTCPP_INCLUDE_DESTINATION} " )
 message( STATUS "BTCPP_UNIT_TESTS:   ${BTCPP_UNIT_TESTS} " )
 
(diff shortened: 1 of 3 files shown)


psyinf:
any news on this?

facontidavide:
thanks!

vincent-hui:
FYI https://github.com/BehaviorTree/BehaviorTree.CPP/pull/926

PR #843, opened 2024-07-02
something build in msvc fixed

kinly:
1. cmake add msvc option /utf-8
2. xml_parsing using filesystem u8path at c++17


facontidavide:
Your changes don't compile in CI. Reopen this once it is fixed

PR #845, opened 2024-07-03
fixed: support utf-8 path xml-file

kinly:
1. added compile version check to support Chinese path xml-file parsing 
3. cmake add msvc /utf-8 options

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,8 @@ else()
     include(cmake/conan_build.cmake)
 endif()
 
+add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
+add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
 
 #############################################################
 # LIBRARY
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,9 +74,6 @@ else()
     include(cmake/conan_build.cmake)
 endif()
 
-add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -74,6 +74,8 @@ else()
     include(cmake/conan_build.cmake)
 endif()
 
+add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
+add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
```
facontidavide, line 78: can you change this to `target_compile_options`

facontidavide (changes requested):

CMakeLists.txt
kinly:
``` diff
@@ -74,9 +74,6 @@ else()
     include(cmake/conan_build.cmake)
 endif()

-add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")  
-add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
-
 #############################################################
 # LIBRARY

@@ -189,6 +186,7 @@ target_compile_definitions(${BTCPP_LIBRARY} PUBLIC BTCPP_LIBRARY_VERSION="${CMAK
 target_compile_features(${BTCPP_LIBRARY} PUBLIC cxx_std_17)

 if(MSVC)
+       target_compile_options(${BTCPP_LIBRARY} PRIVATE "/source-charset:utf-8")
 else()
     target_compile_options(${BTCPP_LIBRARY} PRIVATE -Wall -Wextra)
 endif()
```
good advice, has changed

PR #855, opened 2024-08-05
WIP: Added a new cmake_build for generating a debian

tarunkartik:
If you want to generate a standalone behaviortree_cpp debian that you can install and link your project against,
- Clone the BehaviorTreee.CPP repo
- cd into the cloned repo
- run the following commands:
```
cmake -S . -B build -DCMAKE_BUILD=ON
cmake --build build/ -j{Number_of_cores} --target package
```
Replace {Number_of_cores} with how many cores you would like to use to build. 

- If built succesfully, it should generate a ```.deb``` file inside the build folder.
- You can install this debian in any container or environment using ```apt install /path/to/.deb```

### DIFF
--- a/cmake/cmake_build.cmake
+++ b/cmake/cmake_build.cmake
@@ -1,0 +1,64 @@
+
+if(BTCPP_GROOT_INTERFACE)
+    find_package(ZeroMQ REQUIRED)
+    list(APPEND BTCPP_EXTRA_LIBRARIES ${ZeroMQ_LIBRARIES})
+    list(APPEND BTCPP_EXTRA_INCLUDE_DIRS ${ZeroMQ_INCLUDE_DIRS})
+    message(STATUS "ZeroMQ_LIBRARIES: ${ZeroMQ_LIBRARIES}")
+endif()
+
+if(BTCPP_SQLITE_LOGGING)
+    find_package(SQLite3 REQUIRED)
+    list(APPEND BTCPP_EXTRA_LIBRARIES ${SQLite3_LIBRARIES})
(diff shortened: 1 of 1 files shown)


facontidavide:
Is this change tested on Windows? We don't want to break it there

tarunkartik:
I have not tested it on windows. I can do that. It looks like the repo already has a pipeline to test against windows. Are there any instructions on how to trigger them?

CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ option(BTCPP_EXAMPLES   "Build tutorials and examples" ON)
 option(BTCPP_UNIT_TESTS "Build the unit tests" ON)
 option(BTCPP_GROOT_INTERFACE "Add Groot2 connection. Requires ZeroMQ" ON)
 option(BTCPP_SQLITE_LOGGING "Add SQLite logging." ON)
+option(CMAKE_BUILD "Build using CMAKE" OFF)
```
facontidavide, line 33: what is the purpose of adding this? Shouldn't it be the default ON?

facontidavide (changes requested):

CMakeLists.txt
tarunkartik: The intent is that the users would turn this ON only when they need standalone debians that do not depend on ROS (catkin/ament). That was my requirement for a project I have been working on where I am using BTCPP with a custom RTOS.
 Maybe the variable could be named 'GENERATE_DEBS' instead of 'CMAKE_BUILD', if that makes it clearer?

facontidavide:
not interested in supporting this directly

Issue #861, opened 2024-08-26
Bugprone Tick time monitoring


Petzep:
**Describe the bug**
The time measurement of a tick is bug-prone.
https://github.com/BehaviorTree/BehaviorTree.CPP/blob/2a8a226fbbd99f524f0796e0d8a3145773c61c06/src/tree_node.cpp#L109-L111 

This way of measuring time can be reordered by the compiler and will result in a time difference of 0.
A good explanation of this issue can be found here: 
[Enforcing statement order in C++](https://stackoverflow.com/questions/37786547/enforcing-statement-order-in-c/38025837#38025837)

The article also discusses a fix.

And I've also seen examples where they fix it with an atomic fence (which cost's no extra cylces)
https://stackoverflow.com/questions/22278963/how-to-prevent-timestamp-from-being-reorder-out

facontidavide:
you are welcome to submit a PR to address this.

Aglargil:
https://github.com/BehaviorTree/BehaviorTree.CPP/pull/862
This might help?

facontidavide:
I added some changes that should help with this

PR #862, opened 2024-09-06
Refactored the TreeNode::executeTick() function to use a scoped timer…

Aglargil:
… for performance monitoring. (#861)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,6 @@ else()
     include(cmake/conan_build.cmake)
 endif()
 
-
 #############################################################
 # LIBRARY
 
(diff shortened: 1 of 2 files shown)


src/tree_node.cpp
```diff
@@ -106,13 +106,17 @@ NodeStatus TreeNode::executeTick()
     if(!substituted)
     {
       using namespace std::chrono;
+
       auto t1 = steady_clock::now();
+      std::shared_ptr<void> timer(nullptr, [&](...) {
```
facontidavide, line 111:
```suggestion
      // trick to prevent the compile from reordering the order of execution. See #861 
      // This makes sure that the code is executed at the end of this scope
      std::shared_ptr<void> execute_later(nullptr, [&](...) {
```

facontidavide (changes requested):

Issue #851, opened 2024-07-26
No Input/output port order anymore in writeTreeNodesModelXML ?


RhysMcK:
Hi,

I notice in this commit d5ea17b5184feacac3f3763de0730713384bafc9 you removed the input/output port ordering logic. As a result, when using `writeTreeNodesModelXML() ` to auto-generate the `<TreeNodesModel>` tags for Groot2, the ports are in some ?random? order. This is than obviously reflected in Groot2 and not ideal.  What was the rational behind removing this?

![image](https://github.com/user-attachments/assets/756e1e4a-de2d-45c5-a992-089d70c3d0f8)

Thanks very much!

facontidavide:
The correct solution, in my opinion, is to presere the order of the ports as it was described in **providedPorts**.

Unfortunately, this means that we need to apply a change that breaks the API / ABI. 

In the meantimes, can you test the changes here: #870

PR #870, opened 2024-09-21
preserve order of ports in PortsList. Solves issue #851

facontidavide:
This check is probably NOT binary compatible, nor API compatible


facontidavide:
Unfortunately, if I merge this, I need to create BT.CPP 5.0, because EVERYONE will be obliged to recompile their applications.

Still unsure what to do

PR #873, opened 2024-09-24
Support vector<Any> -> vector<typename T::value_type> conversion

dyackzan:
Don't check port type alignment for vector<Any>

### DIFF
--- a/include/behaviortree_cpp/tree_node.h
+++ b/include/behaviortree_cpp/tree_node.h
@@ -30,6 +30,12 @@
 
 namespace BT
 {
+// Helper trait to check if a type is a std::vector
+template <typename T>
+struct is_vector : std::false_type {};
+
+template <typename T, typename A>
+struct is_vector<std::vector<T, A>> : std::true_type {};
 
 /// This information is used mostly by the XMLParser.
 struct TreeNodeManifest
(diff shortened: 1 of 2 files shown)


include/behaviortree_cpp/blackboard.h
```diff
@@ -25,6 +26,19 @@ struct StampedValue
...
 
+// Helper trait to check if templated type is a std::vector
+template <typename T>
+struct is_vector : std::false_type {};
+
+template <typename T, typename A>
+struct is_vector<std::vector<T, A>> : std::true_type {};
+
+// Helper function to check if a demangled type string is a std::vector<..>
+inline bool isVector(std::string type_name)
```
nbbrooks, line 37:
```suggestion
inline bool isVector(const std::string& type_name)
```

nbbrooks (commented):
Could you document the gcc and clang demangled output for the `std::vector<std::string>` for reference here in the PR if it isn't possible to add a unit test establishing the demangled output passes the regex max? Ideally we'd have a unit test for partial specialization as well.

dyackzan:
I checked clang and gcc with the type demangler and get the the same output for the demangled typeid of a `std::vector<string>` for both:
```
std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >
```

nbbrooks (approved):

nbbrooks:
Oops, when I hit `approve` I lost track of my tabs and thought I was approving a submodule update to point to this PR branch :facepalm:

facontidavide:
looks like a good addition. Can you please:

- use pre-commit to fix the formatting?
- add a simple unit test too?

dyackzan:
How does that look @facontidavide ?

dyackzan:
Tagging @facontidavide , could you give me another review on this?

facontidavide:
Ok, I am finally looking into this.
```cpp
  std::string xml_txt_bad = R"(
    <root BTCPP_format="4" >
      <BehaviorTree>
        <Sequence name="root_sequence">
          <OutputVectorStringNode/>
          <InputVectorDoubleNode double_vector="{string_vector}"/>
        </Sequence>
      </BehaviorTree>
    </root>)";
    
ASSERT_NO_THROW(tree = factory.createTreeFromText(xml_txt_bad));
```
In my opinion this is SUPPOSED to throw. you are basically bypassing the entire type safety mechanism...

facontidavide:
Another problem I see is that this will not play nicely with this:

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/examples/t13_access_by_ref.cpp

If I try to access any vector by reference, I will find a `vector<Any>`. This will break other people code.

facontidavide:
Summary: 
- you basically bypass the entire type safety checks for any` vector<T>`.
- accessing the blackboard by reference will return an unexpected type.
- the unit test doesn't seem meaningful to me, in terms of desired functionality. What you want to show is a generic node that can access an existing port as `vector<Any>`

I understand your problem and the annoyance of templates, and I would like to find a solution, but I am not sure that this is the one I want to merge.

I was thinking about doing the conversion to `vector<Any>` lazily, for instance you have your input port **AnyVector** and it should be able to connect and read from` vector<string>`. The problem is when writing back into the blackboard, because we don't want to change the original type.

A potential solution might be to store a to/from `vector<Any>` "converters" and run them only when requested, to avoid altering the original type in the blackboard.

On the other hand, what the "right" type of the blackboard is is based on some "fuzzy" criteria, i.e. the order of the XMl parsing...

I am not sure how to proceed. Probably creating a better test/use case first and then discussing that.

facontidavide:
If you want, we can arrange a meeting, and iterate faster to find alternatives. You can contact me at dfaconti@aurynrobotics.com

PR #874, opened 2024-09-27
Put minitrace in the build_interface

ahoarau:
Building behavior tree with standard cmake (no ament, no conan etc) will result in:

```
CMake Error: install(EXPORT "behaviortree_cppTargets" ...) includes target "behaviortree_cpp" which requires target "minitrace" that is not in any export set.
```
By putting the `minitrace` library link in the build_interface, it allows to pass the configuration and get the final Config.cmake and Targets.cmake

Before:
image
After:
Screenshot 2024-09-27 081358


facontidavide:
thanks

PR #882, opened 2024-11-14
use relative path in .Doxyfile

m-tartari:
This micro PR removes refercences to @facontidavide local environment from the `.Doxyfile` by favoring relative path usage.


lochanRV (changes requested):
I want to contribute to your project?
I am good with cpp and building LLM and rag models

PR #885, opened 2024-11-20
Additional XML verification for ReactiveSequence nodes

### LINKED ISSUE #875: Using a ReactiveSequence with asynch condition
TLDR: don't use an asynch action for the condition in a ReactiveSequence.

I have a simple ReactiveSequence like this. It's very similar to the example [here](https://www.behaviortree.dev/docs/tutorial-basics/tutorial_04_sequence):

```
    <!-- Keep running DummyStatefulAction as long as the monitors are good -->
    <ReactiveSequence>
        <CheckJointVelocityMonitor1/>  <!-- Inherits from BT::RosServiceNode -->
        <DummyStatefulAction/>            <!-- A long-running asynch action -->
    </ReactiveSequence>
```

The prints when this BT is executing are:

```
CheckJointVelocityMonitor1 returning SUCCESS  // This BT node always returns success
Starting DummyStatefulAction                                // This is expected, it comes from .onStart(), which returns BT::NodeStatus::RUNNING
DummyStatefulAction halted!                                  // I don't understand where the halt comes from
CheckJointVelocityMonitor1 returning SUCCESS
Starting DummyStatefulAction
DummyStatefulAction halted!
CheckJointVelocityMonitor1 returning SUCCESS
Starting DummyStatefulAction
DummyStatefulAction halted!
CheckJointVelocityMonitor1 returning SUCCESS
Starting DummyStatefulAction
DummyStatefulAction halted!
  ... loop continues
```

`CheckJointVelocityMonitor1` always returns SUCCESS. I've verified with print statements. `DummyStatefulAction` appears to be ticked once and returns RUNNING, then it gets halted. I don't understand why it gets halted. (Edit: it gets halted because it returns RUNNING.)

Here's the declaration for `DummyStatefulAction`, which inherits from `BT::StatefulActionNode`.

```
class DummyStatefulAction : public BT::StatefulActionNode
{
public:
  DummyStatefulAction(const std::string& name, const BT::NodeConfiguration& config) : StatefulActionNode(name, config)
  {
  }

  static BT::PortsList providedPorts();

  BT::NodeStatus onStart() override;

  BT::NodeStatus onRunning() override;

  void onHalted() override;
};
```

Here's the definition for `DummyStatefulAction`. It only returns RUNNING, never SUCCESS or FAILURE.

```
BT::NodeStatus DummyStatefulAction::onStart()
{
  std::cerr << "Starting DummyStatefulAction" << std::endl;
  return BT::NodeStatus::RUNNING;
}

BT::NodeStatus DummyStatefulAction::onRunning()
{
  std::cerr << "Running DummyStatefulAction" << std::endl;
  return BT::NodeStatus::RUNNING;
}

void DummyStatefulAction::onHalted()
{
  std::cerr << "DummyStatefulAction halted!" << std::endl;
}
```

AndyZe:
Fixes #875

### DIFF
--- a/src/xml_parsing.cpp
+++ b/src/xml_parsing.cpp
@@ -549,7 +549,7 @@ void VerifyXML(const std::string& xml_text,
           {
             const std::string child_name = node->FirstChildElement()->Name();
             const auto child_search = registered_nodes.find(child_name);
-            auto child_type = child_search->second;
+            const auto child_type = child_search->second;
             if(child_type == NodeType::CONTROL &&
                ((child_name == "ThreadedAction") ||
                 (child_name == "StatefulActionNode") ||
--- a/src/xml_parsing.cpp
+++ b/src/xml_parsing.cpp
@@ -541,6 +541,18 @@ void VerifyXML(const std::string& xml_text,
           ThrowError(line_number,
                      std::string("The node <") + name + "> must have 1 or more children");
         }
+        if (name == "ReactiveSequence")
+        {
+          const std::string child_name = node->FirstChildElement()->Name();
+          const auto child_search = registered_nodes.find(child_name);
+          auto child_type = child_search->second;
+          // TODO: what this really needs to check is if the child is async
+          if (child_type != NodeType::CONDITION)
+          {
+            ThrowError(line_number,
+                      std::string("The first child of a ReactiveSequence cannot be asynchronous"));
+          }
+        }
       }
     }
     //recursion
--- a/src/xml_parsing.cpp
+++ b/src/xml_parsing.cpp
@@ -541,16 +541,17 @@ void VerifyXML(const std::string& xml_text,
           ThrowError(line_number,
                      std::string("The node <") + name + "> must have 1 or more children");
         }
-        if (name == "ReactiveSequence")
+        if(name == "ReactiveSequence")
         {
           const std::string child_name = node->FirstChildElement()->Name();
           const auto child_search = registered_nodes.find(child_name);
           auto child_type = child_search->second;
-          // TODO: what this really needs to check is if the child is async
-          if (child_type != NodeType::CONDITION)
+          if(child_type == NodeType::CONTROL &&
+             ((child_name == "ThreadedAction") || (child_name == "StatefulActionNode") ||
+              (child_name == "CoroActionNode")))
           {
-            ThrowError(line_number,
-                      std::string("The first child of a ReactiveSequence cannot be asynchronous"));
+            ThrowError(line_number, std::string("The first child of a ReactiveSequence "
+                                                "cannot be asynchronous"));
           }
         }
       }
--- a/src/xml_parsing.cpp
+++ b/src/xml_parsing.cpp
@@ -543,15 +543,25 @@ void VerifyXML(const std::string& xml_text,
         }
         if(name == "ReactiveSequence")
         {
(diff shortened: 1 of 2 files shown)


src/xml_parsing.cpp
```diff
@@ -536,6 +536,25 @@ void VerifyXML(const std::string& xml_text,
       }
       else if(search->second == NodeType::CONTROL)
       {
+        if (name == "Sequence" && children_count <= 2)
+        {
+          ThrowError(line_number,
+                     std::string("A Sequence node must have 2 or more children"));
```
AndyZe, line 542: There's no point to a Sequence with just 1 child

src/xml_parsing.cpp
```diff
@@ -536,6 +536,25 @@ void VerifyXML(const std::string& xml_text,
...
+          ThrowError(line_number,
+                     std::string("A ReactiveSequence node must have 2 or more children"));
+        }
+        if (name == "ReactiveSequence")
+        {
+          auto first_child_node = search->second->FirstChildElement();
+          if (first_child_node != NodeType::CONDITION)
+          {
+            ThrowError(line_number,
+                      std::string("The first child of a ReactiveSequence should be of Condition type"));
```
AndyZe, line 555: The first child of a ReactiveSequence should be a Condition that returns SUCCESS or FAILURE immediately.

src/xml_parsing.cpp
AndyZe: nevermind, this is useful for the first Control node in the tree, I guess. Will revert.

src/xml_parsing.cpp
```diff
@@ -541,6 +541,21 @@ void VerifyXML(const std::string& xml_text,
...
+        if (name == "ReactiveSequence")
+        {
+          const std::string child_name = node->FirstChildElement()->Name();
+          const auto child_search = registered_nodes.find(child_name);
+          auto child_type = child_search->second;
+          if (child_type == NodeType::CONTROL &&
+              ((child_name == "ThreadedAction") ||
+              (child_name == "StatefulActionNode") ||
+              (child_name == "CoroActionNode"))
+             )
```
AndyZe, line 553: @facontidavide this is the best i can come up with to determine if a node is asynch. Am I missing an easier way?

facontidavide: of course, an entire subtree could be asynchronous... but as a first attempt to prevent people from shooting themselves in the foot, it is OK.

AndyZe: Great. I added a unit test. I'll work on expanding the functionality to count # of children that are asynchronous now.

AndyZe: Alright, ready for review I think.

src/xml_parsing.cpp
```diff
@@ -541,6 +541,29 @@ void VerifyXML(const std::string& xml_text,
...
         }
+        if(name == "ReactiveSequence")
+        {
+          size_t async_count = 0;
+          for(auto child = node->FirstChildElement(); child != nullptr;
+              child = child->NextSiblingElement())
+          {
+            const std::string child_name = node->FirstChildElement()->Name();
+            const auto child_search = registered_nodes.find(child_name);
+            auto child_type = child_search->second;
```
AndyZe, line 552:
```suggestion
            const auto child_type = child_search->second;
```

facontidavide (approved):

facontidavide:
thanks

PR #887, opened 2024-11-25
Expose return value of wait_for

tonynajjar:


facontidavide:
Thanks. can you modify the documentation of the method to reflect this change?

tonynajjar:
Yep. Done

PR #894, opened 2024-12-10
Allow BT::AnyTypeAllowed as port type and setOuput using BT::Any

### LINKED ISSUE #893: TreeNode::setOutput doesn't consider BT::AnyTypeAllowed when checking for variant port type
I would like to change [this line](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/f5e98fd751523fd2797e5353cb9f1cc765587cb4/include/behaviortree_cpp/tree_node.h#L586) to also consider [`BT::AnyTypeAllowed`](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/f5e98fd751523fd2797e5353cb9f1cc765587cb4/include/behaviortree_cpp/basic_types.h#L87). This would allow the user to correctly allow all types for a specific port. 

When using `BT::Any` instead, the `BT::TypeInfo` of the port defines the functor for converting from string like [here](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/f5e98fd751523fd2797e5353cb9f1cc765587cb4/include/behaviortree_cpp/basic_types.h#L224). This essentially breaks the scripting language because `BT::convertFromString<BT::Any>()` is illegal but will be called internally when a script is evaluated for a variable associated with a `BT::Any` port (This association is defined during tree initialization to validate types).

Therefore, I suggest to replace
```cpp
if constexpr(std::is_same_v<BT::Any, T>)
{
  if(config().manifest->ports.at(key).type() != typeid(BT::Any))
  {
    throw LogicError("setOutput<Any> is not allowed, unless the port "
                     "was declared using OutputPort<Any>");
  }
}
```
with this
```cpp
if constexpr(std::is_same_v<BT::Any, T>)
{
  const std::type_index type = config().manifest->ports.at(key).type();
  if(type != typeid(BT::Any) && type != typeid(BT::AnyTypeAllowed))
  {
    throw LogicError("setOutput<BT::Any> is not allowed, unless the port "
                     "was declared using OutputPort<BT::Any> or "
                     "OutputPort<BT::AnyTypeAllowed>");
  }
}
```
from [here](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/f5e98fd751523fd2797e5353cb9f1cc765587cb4/include/behaviortree_cpp/tree_node.h#L586-L593).

It would be nice if what I'm trying to do here would be possible with `BT::Any` alone but I guess this requires to dig deep in the type validation and scripting logic.

robin-mueller:
This pull request enables to use `BT::AnyTypeAllowed` when specifying the type of a port, basically meaning that this port has no static type and the node will determine the type of the variable dynamically.

See #893

### DIFF
--- a/include/behaviortree_cpp/tree_node.h
+++ b/include/behaviortree_cpp/tree_node.h
@@ -585,10 +585,12 @@ inline Result TreeNode::setOutput(const std::string& key, const T& value)
 
   if constexpr(std::is_same_v<BT::Any, T>)
   {
-    if(config().manifest->ports.at(key).type() != typeid(BT::Any))
+    const std::type_index type = config().manifest->ports.at(key).type();
+    if(type != typeid(BT::Any) && type != typeid(BT::AnyTypeAllowed))
     {
-      throw LogicError("setOutput<Any> is not allowed, unless the port "
-                       "was declared using OutputPort<Any>");
+      throw LogicError("setOutput<BT::Any> is not allowed, unless the port "
+                       "was declared using OutputPort<BT::Any> or "
+                       "OutputPort<BT::AnyTypeAllowed>");
     }
   }
 


facontidavide:
thanks for the suggestion. I wonder how this should be documented in the tutorials. 

What will it be the expected difference when usinf **Any** or **AnyTypeAllowed**?

facontidavide:
change already applied. thanks!!!

Issue #832, opened 2024-06-05
script compare with  negative number exception


jiezenghuang:
example as below：

`static const char* xml_text = R"(
 <root BTCPP_format="4">
     <BehaviorTree>
        <Sequence>
            <Script code=" A:=0 " />
            <AlwaysSuccess _failureIf="A!=-1"/>
        </Sequence>
     </BehaviorTree>
 </root>
 )";

int main(int argc, char** argv)
{
    BT::BehaviorTreeFactory factory;

    try
    {
        auto tree = factory.createTreeFromText(xml_text);
        tree.tickWhileRunning();
    }
    catch(const std::exception& e)
    {
        std::cerr << e.what() << '\n';
    }  

    return 0;    
}`

output：
Error in the script "A!=-1"
error: while parsing Grammar::Expression
     |
   1 | A!=-1
     |    ^ operator cannot be mixed with previous operators

Aglargil:
@facontidavide 
https://github.com/BehaviorTree/BehaviorTree.CPP/pull/896
The content of this PR may resolve this issue.

PR #896, opened 2024-12-12
Fix script compare with negative number exception

Aglargil:
# Fix issue:
#832

# Analysis:
In the example of nested groups in Lexy, the matching behavior is somewhat strange. Therefore, it is recommended to temporarily avoid using nested groups.
https://github.com/foonathan/lexy/issues/219

# Solution:
Remove nested groups and move `string_concat` into comparison operations. This actually aligns better with intuition, since theoretically it should be possible to perform `comparison` operations between two `string_concat `operations.

# References:
https://github.com/foonathan/lexy/blob/77ba5fa5a62f16b3ab5440c6d6ef42a7cdd176bb/tests/lexy/dsl/expression.cpp#L1193-L1196
https://lexy.foonathan.net/reference/dsl/expression/#groups

### DIFF
--- a/examples/t09_scripting.cpp
+++ b/examples/t09_scripting.cpp
@@ -10,7 +10,7 @@ static const char* xml_text = R"(
         <Sequence>
             <Script code=" msg:='hello world' " />
             <Script code=" A:=THE_ANSWER; B:=3.14; color:=RED " />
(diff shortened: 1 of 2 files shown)


facontidavide:
thanks!

PR #907, opened 2024-12-30
Adds "Other Attributes" to a Tree Node's configuration with contains non-port fields

dsobek:
This parses and stores xml attributes in Tree nodes that are not ports so that a client can retrieve these fields from the node's config.


facontidavide:
closing in favor of #910

PR #908, opened 2025-01-03
Replace the while loop with an if statement because placing a while s…

lucvt001:
…tatement is against the spirit of asynchronous actions and will cause things such as RosSubscriber to behave incorrectly


facontidavide:
I think that this PR is full of undesired changes. Clean it up before submitting it again

PR #913, opened 2025-01-21
Fix minitrace export when included via fetchcontent

psyinf:
When importing BehaviorTree.CPP via fetchcontent/cpm a CMake configuration error is issued
"CMake error : install(EXPORT "behaviortree_cppTargets" ...) includes target "behaviortree_cpp" which requires target "minitrace" that is not in any export set."

Implicitly adding minitrace to the install targets fixes this


psyinf:
> weird, this should be a private dependency.

You are right. I'm seeing this using CMake 3.30.3. I'll check with a newer version and if the problem persists I'll create a minimal example

psyinf:
Ok, found the problem: This was fixed with 2bc72fdbf67bae7a4561cc0cea50b723d5acb672. Unfortunately I use the last tag [4.6.2](https://github.com/BehaviorTree/BehaviorTree.CPP/releases/tag/4.6.2) where this issue is still present. 
I'll close the PR. Any plans on a new release?

PR #914, opened 2025-01-28
Fix/use correct compiler pixi/conda

tony-p:
Use the correct compiler in the `pixi.toml` which removes the need for some conda specific things I put in originally, will also automatically select correct compiler for the OS.

Also bring version more in line with conan requirements, and update pixi versions in pipeline.

### DIFF
--- a/.github/workflows/pixi.yaml
+++ b/.github/workflows/pixi.yaml
@@ -13,9 +13,9 @@ jobs:
     steps:
       # Pixi is the tool used to create/manage conda environment
       - uses: actions/checkout@v3
-      - uses: prefix-dev/setup-pixi@v0.4.1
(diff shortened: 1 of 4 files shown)


facontidavide:
Just a question for a person unfamiliar with pixi. Should the **lock** file be there?

tony-p:
Typically yes, it ensures reproducibility, especially when you are trying to go back in time to an old version.

So the toml file describes the constraints, and then solver finds a valid set and fixes them in the lock file.

the lock file will be updated if you change the constraints, and if you want to try bumping all the versions, you can remove the lock file and let it regenerate.

The default parameters of the setup-pixi action will actually ensure that the committed lock file is up to date with the constraints.

PR #915, opened 2025-01-29
Add type & value fields to primitive json types

dyackzan:
This aligns with how custom types are represented by the JsonExporter.

* Update both toJson & fromJson functions in the JsonExporter
* Add & update tests


facontidavide:
1. Is this change back-compatible with the previous JSON format?
2. Can we opt-out if people don't want to use this format?

dyackzan:
1. If I understand your meaning correctly then I don't think this is backward compatible: if someone has a JSON serialization of their blackboard with generic types on it from before this change and they try to load it using `exporter.fromJson` with this change then it will fail and give them the message: "Missing fileld '__type'." I had to update the tests to account for this.
2. I did not implement a way to opt out of this but I believe this way makes more sense for everything to have a __type field rather than just custom types

Are you thinking you don't want to release this in a non-breaking release if there is no way to opt out?

PR #925, opened 2025-02-03
feat: add fuzzing harnesses

cktii:
I still have a bunch of other crashes, esp. for the type-system and casting. I figured why not try to up-stream the fuzzing efforts to have it saved in a central place.

While we're at it, I'd like to add this project to [OSS-Fuzz](https://github.com/google/oss-fuzz) and would prepare an appropriate PR for this next. If it gets approved, it would mean continuous fuzzing/testing for BehaviorTree.CPP and bug reports would come your way without e.g. me having to file anything.

That said, to get that one merged over there, I will need your approval from one of the maintainers here. Eying @facontidavide or @miccol. What do you think? Would that be something you guys would be interested in?


cktii:
This should work now and hopefully pass all the test, once this gets merged I can open a PR (which is ready locally) on the OSS-Fuzz side of things

cktii:
@facontidavide, I thought I had the pre-commit fixed, looks like I forgot to push the changes to the corpus files. Please re-run the workflow, it should be all good now :).

Also, if you're about to merge this, drop a comment in google/oss-fuzz#13008 to drive this effort forward, thanks!

cktii:
@facontidavide kind reminder to get this merged, Google approved of the integration and its just a matter of getting this PR in now and you dropping an `OK` in the linked PR

facontidavide:
This honestly add too much noise to the current **CMakeList.txt**.
Can you please move some of this logic in a separate **.cmake** file?

cktii:
Yeah sure thats actually what I should have done. I'll push something for that in a bit!

facontidavide:
thanks!!!

Issue #921, opened 2025-01-31
Null pointer dereference in ImportBlackboardFromJSON when entry creation fails


cktii:
## Description
A null pointer dereference occurs in `ImportBlackboardFromJSON` when attempting to use an `entry` pointer without verifying that the entry was successfully created. This happens because the code assumes `createEntry` followed by `getEntry` will always return a valid `entry`.

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/48f6c5bd1d797c73a562f1c92f404d0bb944e018/src/blackboard.cpp#L280-L295

Found in commit: https://github.com/BehaviorTree/BehaviorTree.CPP/commit/48f6c5bd1d797c73a562f1c92f404d0bb944e018

## Bug Class
Memory Safety - Null Pointer Dereference

## Root Cause

```cpp
void ImportBlackboardFromJSON(const nlohmann::json& json, Blackboard& blackboard)
{
  for(auto it = json.begin(); it != json.end(); ++it)
  {
    if(auto res = JsonExporter::get().fromJson(it.value()))
    {
      auto entry = blackboard.getEntry(it.key());
      if(!entry)
      {
        blackboard.createEntry(it.key(), res->second);
        entry = blackboard.getEntry(it.key());  // Could still return nullptr
      }
      entry->value = res->first;  // CRASH: Null pointer dereference
    }
  }
}
```

The issue occurs because:

- Code checks if initial `getEntry` returns `nullptr`
- Attempts to create `entry` if it doesn't exist
- Gets `entry` again but doesn't verify it's valid
- Dereferences the `entry` pointer without checking

### GDB

```gdb
pwndbg> p entry->value
Cannot access memory at address 0x0
pwndbg> p entry
$1 = std::shared_ptr<BT::Blackboard::Entry> (empty) = {
  get() = 0x0
}
pwndbg> p res->first
$2 = {
  _any = {
    storage = {
      dynamic = 0xffffffffe010006e,
      stack = {
        __data = "n\000\020\340\377\377\377\377\000\000\000\000\000\000\000",
        __align = {<No data fields>}
      }
    },
    vtable = 0x555556412f40 <linb::any::vtable_for_type<long>()::table>
  },
  _original_type = {
    _M_target = 0x7ffff7c6e610 <typeinfo for long>
  }
}
pwndbg> p json
$1 = (const nlohmann::json_abi_v3_11_3::json &) @0x7ffff550dc20: {
[...]

PR #929, opened 2025-02-06
Use recursive call in Blackboard::createEntry

Aglargil:
fix: #921

Modified the implementation of Blackboard::createEntry to use a recursive call to rootBlackboard()->createEntry instead of rootBlackboard()->createEntryImpl.


facontidavide:
Hi,

this seems to be a rather illegal syntax. i believe it should fail, if used.
let me address this differently

facontidavide:
see c23f8ed11ed9fb3b320e9e66848285701ff55285

Issue #939, opened 2025-02-25
Test failures when building with clang


dsobek:
**Describe the bug**
When building with clang on master, tests are failing with c++ exception:
`C++ exception with description "Value outside numeric limits" thrown in the test body.`

`Any.Cast` appears to be the most minimal testcase that is failing. It comes down to casts from any integral type to a double type.

**How to Reproduce**

Please provide a specific description of how to reproduce the issue or source code that can be compiled and executed. Please attach a file/project that is easy to compile, don't copy and paste code snippets!

From project root:
```
CC=clang CXX=clang++ colcon build
./build/behaviortree_cpp/tests/behaviortree_cpp_test
```

**Root cause**

I've found that the root cause is in [checkTruncation](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/a7c90872ee776c840044e49e260ecde69bd69fab/include/behaviortree_cpp/utils/convert_impl.hpp#L90)

In particular if `To` has a larger max value than `From` (say for example `To` is `double` and `From` is `int`),  `static_cast<From>(std::numeric_limits<To>::max())` does not always evaluate to the max value of `From`. In clang, with any optimization (I see behaviortree.cpp is built with `-O3`), this evaluates to 0. I am not sure if this is meant to be undefined behavior but it sure is acting like it.

dsobek:
Potential solution in https://github.com/BehaviorTree/BehaviorTree.CPP/pull/940

Aglargil:
This might help too #932

facontidavide:
I am aware of the issue.

facontidavide:
the change in the code was introduced mased on #919 , as suggested by @cktii

but clearly that is not a solution that works on all platforms.

@Aglargil how is #932 related?

@dsobek your PR #940 doesn't pass the tests in wondows

Aglargil:
> [@Aglargil](https://github.com/Aglargil) how is [#932](https://github.com/BehaviorTree/BehaviorTree.CPP/pull/932) related?

@facontidavide 
I found this error when I was adding the _onStart precondition function to the script, so I fixed it, you can take a look at this commit: https://github.com/BehaviorTree/BehaviorTree.CPP/commit/d8ca1f1b055aaf50d4c95b26f204e787d77b6b71

facontidavide:
fixed with #941 .Thanks @Aglargil you were right!

dsobek:
Thanks for fixing!

[...]

PR #940, opened 2025-02-25
Fix clang test failures from Any.cast

dsobek:
Fixes https://github.com/BehaviorTree/BehaviorTree.CPP/issues/939
Only results in one less detailed error message (but otherwise the same failure conditions are still covered) and fixes the clang build.

To test build in clang and run unit tests. They should now pass.

```
CC=clang CXX=clang++ colcon build
./build/behaviortree_cpp/tests/behaviortree_cpp_test
```


facontidavide:
thanks, but I think that I have found a better solution. See #941

Issue #943, opened 2025-03-01
Allow passing a derived object to a port requiring a base class


captain-yoshi:
**Is your feature request related to a problem? Please describe.**
There is a lot of boilerplate code needed for converting libraries, relying on inheritance, into BT nodes. Consider this example:
```cpp
// Base class
class Greeter {
 public:
  virtual std::string show_msg() const = 0;
  virtual ~Greeter() {}
};

// Derived class
class HelloGreeter : public Greeter {
 public:
  std::string show_msg() const override { return "hello"; }
  void setDerivedParameter(int n);
};
```
I have 2 options if I want to use a unique method from the derived class as well as store it's base class somewhere, e.g. inside a vector.
```xml
<!-- OPTION 1- return base and derived class -->
<CreateHelloGreeter out_derived="{hello_greeter}" out_base="{greeter}" />

<!-- OPTION 2- convert to base -->
<CreateHelloGreeter out_derived="{hello_greeter}" />
<UpcastHelloGreeter in={hello_greater} out_base="{greeter}" />

<setDerivedParameter in_derived="{hello_greater}" n="2" />
<VectorInsertGreeter in_base="{greeter}" />
```

**Describe the solution you'd like**
I would like the ability to allow passing a derived object directly to a port requiring a base class.
```xml
<CreateHelloGreeter out_derived="{hello_greeter}" />

<setDerivedParameter in_derived="{hello_greater}" n="2" />
<InsertToGreeterVector in_base="{hello_greeter}" />
```
As you see, this would reduce boilerplate code in the C++ and in the XML representation. Is there any way to reduce the strictness of the custom `std::any` used in this project to enable this feature ? I think that this project https://github.com/samee/any_of implements this behavior.

There is also the problem that BT evaluates ports type for consistency. I don't think one can deduce the base class of a derived class 
 `typeid`...

facontidavide:
this is a good point.

Yes, I think it is safe to reduce the strictness of the any_cast.

Can you please help creating a PR with a unit test of the desired behavior, i.e. somehting like this should work out of the box


```xml
<CreateHelloGreeter out_derived="{hello_greeter}" />
<setDerivedParameter in_derived="{hello_greater}" n="2" />
<ShowGreetmessage in_base="{hello_greeter}" />
```

PR #944, opened 2025-03-01
Allow upcasting

captain-yoshi:
Unit tests with desired behavior from #943. 

### Any Tests Failures
```bash
/home/captain-yoshi/ws/ros/mimik_ws/src/BehaviorTree.CPP/tests/gtest_any.cpp:271: Failure
Expected: auto res = a.cast<std::shared_ptr<Greeter>>() doesn't throw an exception.
  Actual: it throws.
/home/captain-yoshi/ws/ros/mimik_ws/src/BehaviorTree.CPP/tests/gtest_any.cpp:273: Failure
Value of: a.castPtr<std::shared_ptr<Greeter>>()
  Actual: false
Expected: true
```

### Blackboard Test Failure
```bash
C++ exception with description "The creation of the tree failed because the port [hello_greeter] was initially created with type [std::shared_ptr<HelloGreeter>] and, later type [std::shared_ptr<Greeter>] was used somewhere else." thrown in the test body.
```

### DIFF
--- a/include/behaviortree_cpp/utils/safe_any.hpp
+++ b/include/behaviortree_cpp/utils/safe_any.hpp
@@ -368,31 +368,38 @@ public:
 
         try
         {
+          // Case 1: If Base and Derived are the same, no casting is needed
+          if constexpr(std::is_same_v<RootBase, Derived>)
+          {
+            return linb::any_cast<std::shared_ptr<RootBase>>(&_any);
+          }
+
+          // Case 2: Previously cached
+          if(_cached_type == typeid(T))
+          {
+            return reinterpret_cast<T*>(&_cached_derived_ptr);
+          }
+
           // Attempt to retrieve the stored shared_ptr<Base> from the Any container
           auto base_ptr = linb::any_cast<std::shared_ptr<RootBase>>(&_any);
           if(!base_ptr)
             return nullptr;
 
-          // Case 1: If Base and Derived are the same, no casting is needed
-          if constexpr(std::is_same_v<RootBase, Derived>)
-          {
-            return reinterpret_cast<T*>(base_ptr);
-          }
-
-          // Case 2: Originally stored as shared_ptr<Derived>
+          // Case 3: Stored as Derived originally
           if(_original_type == typeid(std::shared_ptr<Derived>))
           {
             _cached_derived_ptr = std::static_pointer_cast<Derived>(*base_ptr);
-            return reinterpret_cast<T*>(&_cached_derived_ptr);
           }
-
-          // Case 3: Fallback to dynamic cast
-          auto derived_ptr = std::dynamic_pointer_cast<Derived>(*base_ptr);
-          if(derived_ptr)
+          // Case 4: Fallback to dynamic cast
+          else
           {
+            auto derived_ptr = std::dynamic_pointer_cast<Derived>(*base_ptr);
+            if(!derived_ptr)
+              return nullptr;
             _cached_derived_ptr = derived_ptr;
-            return reinterpret_cast<T*>(&_cached_derived_ptr);
           }
+          _cached_type = typeid(T);
+          return reinterpret_cast<T*>(&_cached_derived_ptr);
         }
         catch(...)
         {
@@ -427,6 +434,7 @@ private:
   linb::any _any;
   std::type_index _original_type;
   mutable std::shared_ptr<void> _cached_derived_ptr = nullptr;
+  mutable std::type_index _cached_type = typeid(void);
 
   //----------------------------
 


captain-yoshi:
*There maybe a better way to allow type casting of a polymorphic class.*

​This pull request introduces support for upcasting and downcasting `std::shared_ptr` instances (must be polymorphic and has their base registered) within the `Any` class, facilitating polymorphic type handling. By leveraging type traits, the `Any` class can now perform safe and efficient conversions between shared pointers of related types.
- Introduced `any_cast_base<T>` trait to register base types for polymorphic casting.
- Added `is_shared_ptr` and `IsPolymorphicSharedPtr` traits to detect eligible types.
- Modified `Any` to store shared_ptr as its registered base class if available.
- Enhanced castPtr() and tryCast() to support safe downcasting from stored base class to derived class using static or dynamic pointer casting.
- Ensures shared_ptr<Derived> can be stored and retrieved as shared_ptr<Base> when registered.

So users who wants this behavior needs to register their classes to it's base class. I don't think we can get away without this process...
```cpp
// Register cast base type for self to allow direct cast, otherwise defaults to a dynamic cast
template <>
struct BT::any_cast_base<Greeter>
{
  using type = Greeter;
};

// Register cast base type for HelloGreeter
template <>
struct BT::any_cast_base<HelloGreeter>
{
  using type = Greeter;
};
```
Optionally we could add a macro to simplify this process (not in this PR):
```cpp
#define BT_REGISTER_BASE_TYPE(Derived, Base)    \
  template <>                                   \
  struct BT::any_cast_base<Derived>             \
  {                                             \
    using type = Base;                          \
  };
  
BT_REGISTER_BASE_TYPE(Greeter, Greeter)
BT_REGISTER_BASE_TYPE(HelloGreeter, Greeter) 
```

It's a little bit more permissive, then what I intended initially. You can now downcast to something that is invalid (must still be a valid class of the stored base class), e.g. the blackboard will not see a mismatch, but it will fail when trying to retrieve it in a BT::Node. 

Regarding the previous point, should we try to downcast early, in the blackboard to stop the tree from running early if it is invalid ? Instead of running the tree and let it fail down the line ?

----
The Sonarcube check fails, but it does not seem related to this PR.

captain-yoshi:
The `castPtr` function is a little bit tricky when downcasting...

I would prefer solution 2 or 3.

### Solution 1
Returns a `T::element_type* castPtr()` when T is a shared pointer with a polymorphic class and a registered base. For everything else T* is returned.

| Pros          | Cons                                                                                                                   |
|---------------|------------------------------------------------------------------------------------------------------------------------|
| No complexity | 2 different return types depending on the type.<br>Cannot retrieve the shared pointer, only access to the raw pointer. |

*Implemented in 3021ff0*

### Solution 2
Clean API: `T* castPtr()` for all types.
| Pros       | Cons                                                                                                                                                                       |
|------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Returns T* | Need to store a std_shared\<void> member to return a pointer to a Derived class.<br>Lifetime of pointer not garanteed*, another call to castPtr may change this cached value. <br><br>*If getLockedPortContent is used, there is no problem.|

*Implemented in d11512a*

### Solution 3
Seperate castPtr function when T is a shared pointer with a polymorphic class and a registered base, e.g. `castShared`.
| Pros          | Cons                                                                    |
|---------------|-------------------------------------------------------------------------|
| No complexity | User has 2 different functions depending on what is stored in the _any. |

*Not implemented*

facontidavide:
There is a conflict that need to be resolved. Also, correct me if i am wrong, but isn't this change allowing two ports with derived classes to connect?

For instance, given **Cat <- Animal** and **Dog <- Animal**, I would be allowed to connect a Cat to a Dog...

captain-yoshi:
> For instance, given Cat <- Animal and Dog <- Animal, I would be allowed to connect a Cat to a Dog...

At the moment yes. If we want to make it fail when creating a Tree node from xml, we would need to store the type id of the base class and the original class, because at this stage, I think the any values are still empty.
<p align="center">

</p>

But trying to have a SphinxCat into a registered Cat port would also fail because only the the SphinxCat and Animal typeid would be stored when registering SphinxCat. We could fix this by storing a vector (or limit with a max depth) of typeid's from current class to root base class to enable this case.

I would also prefer to make it fail early when trying to connect incompatible types. What wold be the way to go ? Only allow a depth of 2 base classes, allow a max range of depth or allow any depth ? Or something else entirely...

captain-yoshi:
This succeeds:
```cpp
std::string xml_txt = R"(
<root BTCPP_format="4" >
  <BehaviorTree ID="Main">
    <Sequence>

      <CreateCat out_cat="{cat}" />
      <CreateAnimal in_animal="{cat}" />
    </Sequence>
  </BehaviorTree>
</root>)";

BehaviorTreeFactory factory;
factory.registerNodeType<CreateAnimal>("CreateAnimal");
factory.registerNodeType<CreateCat>("CreateCat");
factory.registerNodeType<CreateSphynxCat>("CreateSphynxCat");
factory.registerNodeType<CreateDog>("CreateDog");

auto tree = factory.createTreeFromText(xml_txt);

ASSERT_EQ(status, NodeStatus::SUCCESS);
```
and also this succeeds:
```cpp
std::string xml_txt = R"(
<root BTCPP_format="4" >
  <BehaviorTree ID="Main">
    <Sequence>
      <CreateCat out_cat="{cat}" />
      <CreateSphynxCat in_sphynx="{cat}" />
    </Sequence>
  </BehaviorTree>
</root>)";

BehaviorTreeFactory factory;
factory.registerNodeType<CreateAnimal>("CreateAnimal");
factory.registerNodeType<CreateCat>("CreateCat");
factory.registerNodeType<CreateSphynxCat>("CreateSphynxCat");
factory.registerNodeType<CreateDog>("CreateDog");

ASSERT_ANY_THROW(auto tree = factory.createTreeFromText(xml_txt));
```

but this should throw at creation but succeeds.
```cpp
std::string xml_txt = R"(
<root BTCPP_format="4" >
  <BehaviorTree ID="Main">
    <Sequence>

      <CreateCat out_cat="{cat}" />
      <CreateAnimal out_animal="{cat}" />
      <CreateCat out_cat="{cat}" />
    </Sequence>
  </BehaviorTree>
</root>)";

BehaviorTreeFactory factory;
factory.registerNodeType<CreateAnimal>("CreateAnimal");
factory.registerNodeType<CreateCat>("CreateCat");
factory.registerNodeType<CreateSphynxCat>("CreateSphynxCat");
factory.registerNodeType<CreateDog>("CreateDog");

ASSERT_ANY_THROW(auto tree = factory.createTreeFromText(xml_txt));

// Throws further down in the Blackboard::set of the CreateAnimal
// NodeStatus status = tree.tickWhileRunning();
```

Upcasting from Input port is correct, but upcasting to Output ports must not. But we currently don't have the port direction info, so I cannot preempt early in the createNodeFromXml function:
```cpp
for(const auto& [port_name, port_info] : manifest->ports)
{
  auto remap_it = port_remap.find(port_name);
  if(remap_it == port_remap.end())
  {
    continue;
  }
  StringView remapped_port = remap_it->second;

  if(auto param_res = TreeNode::getRemappedKey(port_name, remapped_port))
  {
    // port_key will contain the key to find the entry in the blackboard
    const auto port_key = static_cast<std::string>(param_res.value());

    // if the entry already exists, check that the type is the same
    if(auto prev_info = blackboard->entryInfo(port_key))
    {
      // Check consistency of types.
      bool port_type_mismatch =
          (prev_info->isStronglyTyped() && port_info.isStronglyTyped() &&
           prev_info->type() != port_info.type());

      // *********************************************************
      // Only allow the next mismatch if the port_info is an input or bidirectionnal ?

      // allow mismatch if the previous type appears in the base_chain
      // of the current port (i.e., valid polymorphic upcast)
      if(port_type_mismatch && !prev_info->baseChain().empty() &&
         !port_info.baseChain().empty() &&
         prev_info->isConvertibleFrom(port_info.baseChain()[0]))
      {
        port_type_mismatch = false;
      }

      // special case related to convertFromString
      bool const string_input = (prev_info->type() == typeid(std::string));

      if(port_type_mismatch && !string_input)
      {
        blackboard->debugMessage();

        throw RuntimeError("The creation of the tree failed because the port [",
                           port_key, "] was initially created with type [",
                           demangle(prev_info->type()), "] and, later type [",
                           demangle(port_info.type()), "] was used somewhere else.");
      }
    }
    else
    {
      // not found, insert for the first time.
      blackboard->createEntry(port_key, port_info);
    }
  }
}
```  
Is the port direction really needed ? If yes how do you propose to have access to this information ?

captain-yoshi:
I have now replaced the Greeter hierarchy with the Animal one (easier too understand).

I added some new tests. **Cat <- Animal** and **Dog <- Animal** now throws when constructing the TreeNode.

----
I think the last thing missing is the serialization of the TreeNodeModel in XML ? I think we would need to store the chain base, e.g. :
```xml
<output_port name="collision_object"
             type="std::shared_ptr&lt;Sphinx&lt;std::allocator&lt;void&gt; &gt; &gt;"
             base_chain="Sphinx;Cat;Animal" />
```

Do we need something for deserializing the base_chain ? Is this even needed, I got the impression that we always need to register the node type in the BehaviorTreeFactory...

facontidavide:
thanks for working on this. 

the proposed solution seems to work but it adds some (necessary) complexity and I need to take my time (something that right now I don't have in abundance, to think about alternatives... if they exist.

Please be patient, i will look at this in the next couple of weeks.

PR #955, opened 2025-04-01
Bug fix/set blackboard

devis12:
SetBlackboard action bug fix

Previous SetBlackboard action node did not update the sequence id, the timestamp when called on a pre-existing entry.
Moreover, it allowed to bypass type checking.

This occurred only when the input was a blackboard pointer.

By using the BT::Blackboard::set method I believe also is thread safe which before was not, but I provided test that were failing previously and now are passing exclusively for the two cases above


facontidavide:
very good work. thanks for the tests!

about the issue you mentioned, I will need to think about it

PR #961, opened 2025-04-15
Fix testing CMake issue to resolve Rolling regression

### LINKED ISSUE #960: Regression on ROS 2 Buildfarm for Rolling
`bwehaviortree_cpp 4.6.2` is failing to build for Rolling on the buildfarm with the error below. This is blocking the [announced Rolling sync](https://discourse.ros.org/t/preparing-for-rolling-sync-2025-04-11/43166). We will proceed with the sync which will remove these binaries from Rolling but they will be added once the build is fixed and a new release is bloomed.

```
23:12:15 -- BTCPP_EXTRA_LIBRARIES: $<BUILD_INTERFACE:ament_index_cpp::ament_index_cpp>;$<BUILD_INTERFACE:/usr/lib/x86_64-linux-gnu/libzmq.so>;$<BUILD_INTERFACE:/usr/lib/x86_64-linux-gnu/libsqlite3.so>
23:12:15 -- BTCPP_LIB_DESTINATION:   lib 
23:12:15 -- BTCPP_INCLUDE_DESTINATION: include 
23:12:15 -- BTCPP_UNIT_TESTS:   ON 
23:12:15 CMake Error at /usr/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
23:12:15   Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR
23:12:15   GTEST_MAIN_LIBRARY)
23:12:15 Call Stack (most recent call first):
23:12:15   /usr/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
23:12:15   /usr/share/cmake-3.28/Modules/FindGTest.cmake:270 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
23:12:15   tests/CMakeLists.txt:60 (find_package)
```

Detailed logs: https://build.ros2.org/view/Rbin_uN64/job/Rbin_uN64__behaviortree_cpp__ubuntu_noble_amd64__binary/61/consoleFull

sarcasticnature:
Hopefully resolves #960 (with a small typo fix thrown in for good measure)


sarcasticnature:
Welp, I screwed something up based on all the failed CI jobs, I'll give this a closer look

sarcasticnature:
I think the issue is that _many_ things were actually using the "non-ros" `else` clause in the testing CMake because they had `BUILD_TESTING` set to `OFF` (even though they actually use the test executable...)

sarcasticnature:
if @facontidavide is ok with it I may just strip out the `BUILD_TESTING` given it wasn't working as expected anyway

CMakeLists.txt
```diff
@@ -55,21 +55,21 @@ if ( ament_cmake_FOUND )
 
     add_definitions( -DUSING_ROS2 )
     message(STATUS "------------------------------------------")
-    message(STATUS "BehaviourTree is being built using AMENT.")
+    message(STATUS "BehaviorTree is being built using AMENT.")
```
Yadunund, line 58: These changes seem unnecessary.

tests/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ set(TEST_DEPENDECIES
     foonathan::lexy
     bt_sample_nodes)
 
-if(ament_cmake_FOUND AND BUILD_TESTING)
+if(ament_cmake_FOUND)
```
Yadunund, line 41: I think the original check is correct

```suggestion
if(ament_cmake_FOUND AND BUILD_TESTING)
```

The problem here is that gtest was being found and linked transitively via `ament_cmake`. We'll likely need to explicitly call `find_package(ament_cmake_gtest REQUIRED)` before we call `ament_add_gtest` below and add a `<test_depend>` for the same in package.xml.

Yadunund (changes requested):

tests/CMakeLists.txt
sarcasticnature: also, as I noted in my comments on the issue, it seems that this case of the if statement isn't even being hit on the build farm

https://github.com/BehaviorTree/BehaviorTree.CPP/issues/960#issuecomment-2803472805

facontidavide:
looks good to me, thanks for addressing this

CMakeLists.txt
facontidavide: I like ti, let's keep it

Issue #930, opened 2025-02-06
Different behavior using `registerSimpleAction()` and loadSubstitutionRuleFromJSON()


siddux:
**Describe the bug**
I've a BT running inside a ROS node. More precisely I tick the tree on the `execute()` function of an action server. Everything worked fine but I wanted to add some testing using the mock nodes as described on the tutorial. Once running with the substitutions, the tree gets stuck right  after the async delay finishes (immediately when enters on the subsituted node if the delay is 0). I've been playing around it so I found that just running it using a multithreaded executor solves the issue (even though every now and then it still gets stuck if I send goals quick enough). 

This wouldn't be a problem by itself but later one I've realized that if I change my rules json file to something like this:

```json
{
  "TestNodeConfigs": {
    
  },

  "SubstitutionRules": {
    "script_2": "MyTest"
  }
}
```

And on my code, I define the node like this:

```c++
factory.registerSimpleAction("DummyAction", [](BT::TreeNode& self){
    std::cout << "DummyAction substituting: "<< self.name() << std::endl;
    return BT::NodeStatus::SUCCESS;
  });
```

It runs perfectly on a single threaded executor. I've been analyzing the 2 methods from the factory that are used and I see the way of register the node is different but I don't have enough knowledge about the library to spot why this happens.

So my question is if that's really a bug or a design decision. In any case it would be nice to either find a way to fix it or to tell the user the limitation of each type of subsitution.

*How to Reproduce**

I'll update this later once I've a cleaner example that can be easily run.

Issue #927, opened 2025-02-04
Mock testing substitutions making the tree hang


pepeRossRobotics:
**Describe the bug**
I am testing simple mock testing using the guide [here](https://www.behaviortree.dev/docs/tutorial-advanced/tutorial_15_replace_rules) and I am experiencing that my tree stops ticking once the tree gets to the substituted node.

The tree is loaded using XML and the substitution rules are loaded using a JSON file

**How to Reproduce**
This is my tree:
```XML
<root BTCPP_format="4">
    <BehaviorTree ID="MainTree">
        <Sequence>
            <PrintToConsole
                name="PrintToConsole"
                string="testing"
                prefix="pre sub"
                timestamp="false"
            />
            <Script name="script_1"  code=" next_inspection_point:='0' "/>

            <PrintToConsole
                name="PrintToConsole"
                string="testing"
                prefix="post sub"
                timestamp="false"
            />
            <Delay delay_msec="2000">
                <AlwaysSuccess/>
            </Delay>
        </Sequence>
    </BehaviorTree>
</root>
```
This is the part of the code that runs the tree:
```cpp
void Mk4BehaviorTrees::execute(const std::shared_ptr<BtGoalHandle> goal_handle)
{
  RCLCPP_INFO(get_logger(), "Executing goal");
  const auto goal_details = goal_handle->get_goal();

  if (does_inspection_file_exist(goal_details->path)) {
    load_inspection(goal_details->path);
  } else {
    send_no_inspection_file(goal_handle, goal_details->path);
    return;
  }

  // This is called if the using_mock_nodes is set to true
  load_mock_nodes();


  start_time_ = get_clock()->now();
  while (rclcpp::ok()) {
    if (goal_handle->is_canceling()) {
      running_tree_->haltTree();
      cleanup_tree();
      send_goal_cancelled(goal_handle);
      is_tree_being_executed_ = false;
      return;
    }

    send_feedback(goal_handle);
    running_tree_->tickOnce();
    RCLCPP_INFO(get_logger(), "Ticking");
    running_tree_->sleep(std::chrono::milliseconds(100));
  }

  if (rclcpp::ok()) {
    cleanup_tree();
    send_goal_completed(goal_handle);
  }
}
```
Without the substitutions this is the trees output:
```
[component_container-1] [INFO] [1738660757.723056057] [behavior_trees_node]: 
[component_container-1] pre sub
[component_container-1] testing
[component_container-1] 
[...]

PR #967, opened 2025-04-23
Fix dangling‐capture in TestNodeConfig::complete_func to eliminate flaky IDLE returns

pepeRossRobotics:
# Summary
When using TestNode for JSON-driven substitution rules, the default complete_func lambda in
TestNodeConfig was capturing this (the address of a temporary). After move-constructing into the member _test_config, that pointer became dangling and often returned IDLE from onRunning(), triggering a LogicError. This patch changes the capture to copy return_status by value, fully eliminating the UB and making the test nodes reliably return SUCCESS or FAILURE as configured.

# Background / Motivation

Reported in issue #927: “TestNode sometimes returns IDLE onRunning()”
Reproduced in a minimal repo: test nodes intermittently returned IDLE even when configured for SUCCESS
Added file/minitrace loggers, saw onRunning() → IDLE and confirmed it came from complete_func()
Debugging revealed `complete_func = [this](){…}` pointed at a destroyed temporary
Changes

```
--- a/include/behaviortree_cpp/test_node.h
+++ b/include/behaviortree_cpp/test_node.h
@@ struct TestNodeConfig
-  std::function<NodeStatus(void)> complete_func = [this]() { return return_status; };
+  /// Capture return_status by value to avoid dangling this-pointer
+  std::function<NodeStatus(void)> complete_func = [ret = return_status]() { return ret; };
```

Before:
```
onStart() → RUNNING
onRunning() → complete_func() returns IDLE  ← UB, crashes
```
After:
```
onStart() → RUNNING
onRunning() → complete_func() returns SUCCESS ← correct
```
# Testing
Added file- and minitrace-logger in user code to capture every tick.
Wrote a small BT with a single <Action ID="Test" …/> substituted via JSON.
Verified 1st tick RUNNING, 2nd tick SUCCESS.
Ran full scenario suite — no more LogicError.

Related issues Issues:
#927: https://github.com/BehaviorTree/BehaviorTree.CPP/issues/927
and maybe
#930:  https://github.com/BehaviorTree/BehaviorTree.CPP/issues/930

Thanks for reviewing 🙏 Let me know if you’d like any tweaks!

### DIFF
--- a/include/behaviortree_cpp/actions/test_node.h
+++ b/include/behaviortree_cpp/actions/test_node.h
@@ -39,7 +39,7 @@ struct TestNodeConfig
 
   /// Function invoked when the action is completed. By default just return [return_status]
   /// Override it to intorduce more comple cases
-  std::function<NodeStatus(void)> complete_func = [this]() { return return_status; };
+  std::function<NodeStatus(void)> complete_func = [ret = return_status]() { return ret; };
 };
 
 /**


facontidavide:
CI is red

facontidavide:
I opted for a different solution, but thanks for finding the issue :smile: 

3540b4a6422975198463653c0aefc7869210063c

facontidavide:
also, thinking more carefully about it, I decided to make a more radical change: 2da790634b748087a5a090f2607c89b208af1cc3

Issue #1003, opened 2025-09-10
Crash in XMLParser::PImpl::loadDocImpl when <include> tag is missing path attribute


ahuo1:
Hi, I found a null pointer vulnerability using my fuzzer.

## Environment
- OS: Ubuntu 22.04
- Compiler: clang 13.0.1
- C++ Standard: C++17
- Sanitizers: AddressSanitizer (ASan) + UndefinedBehaviorSanitizer (UBSan)

## Build Instructions
```
export CXXFLAGS="${CXXFLAGS} -std=c++17 -stdlib=libstdc++ -fsanitize=address -O1 -g"
export CFLAGS="${CFLAGS} -fsanitize=address -O1 -g"
export LDFLAGS="${LDFLAGS} -fsanitize=address"

mkdir build && cd build
cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_FUZZING=ON -DFORCE_STATIC_LINKING=ON
cmake --build . --parallel
```

## Reproduction
Run the fuzzer with a crafted input file:
```
./bt_fuzzer poc
```

## Observed Behavior
Program crashes with ASan/UBSan report:

```
==26672==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f0714d680bc bp 0x7ffc9e93ced0 sp 0x7ffc9e93c688 T0)
==26672==The signal is caused by a READ memory access.
==26672==Hint: address points to the zero page.
    ...
    #5 0x8261dd in std::filesystem::__cxx11::path::path<char const*, std::filesystem::__cxx11::path>(char const* const&, std::filesystem::__cxx11::path::format) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/fs_path.h:289:32
    #6 0x82173f in BT::XMLParser::PImpl::loadDocImpl(tinyxml2::XMLDocument*, bool) /root/behavior/behaviortreecpp/src/xml_parsing.cpp:265:20
    #7 0x82398b in BT::XMLParser::loadFromText(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) /root/behavior/behaviortreecpp/src/xml_parsing.cpp:174:7
[...]

PR #1004, opened 2025-09-10
fix: check path attribute before using

ahuo1:
Fix the null pointer vulnerability in #1003


facontidavide:
can you fix formatting?

ahuo1:
I will reopen a PR :(

Issue #1006, opened 2025-09-10
Stack buffer overflow in ParseScript due to fixed-size error buffer


ahuo1:
Hi, I also found a stack-buffer-overflow vulnerability using my fuzzer.

## Environment
- OS: Ubuntu 22.04
- Compiler: clang 13.0.1
- C++ Standard: C++17
- Sanitizers: AddressSanitizer (ASan) + UndefinedBehaviorSanitizer (UBSan)

## Build Instructions
```
export CXXFLAGS="${CXXFLAGS} -std=c++17 -stdlib=libstdc++ -fsanitize=address -O1 -g"
export CFLAGS="${CFLAGS} -fsanitize=address -O1 -g"
export LDFLAGS="${LDFLAGS} -fsanitize=address"

mkdir build && cd build
cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_FUZZING=ON -DFORCE_STATIC_LINKING=ON -DBUILD_TESTING=OFF
cmake --build . --parallel
```

## Reproduction
Run the fuzzer with a crafted input file:
```
./bt_fuzzer poc
```

## Observed Behavior
Program crashes with ASan/UBSan report:

```
==27299==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffdeab4a080 at pc 0x00000080bef3 bp 0x7ffdeab47c70 sp 0x7ffdeab47c68

    ...
    #24 0x75e930 in BT::ParseScript(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /root/behavior/behavior_fix/src/script_parser.cpp:20:7
[...]

PR #1007, opened 2025-09-10
fix: use dynamically growing error buffer in ParseScript

ahuo1:
fix bug in #1006


ahuo1:
I think this bug also exists in ValidateScript. While fuzzing script_fuzzer, the same overflow was triggered there, so I applied the same fix.

PR #1012, opened 2025-09-26
Improve handling of dependencies

ericriff:
On this PR I've reworked how the 3rd party libraries are consumed to improve the conan support.

The rationale is to give the consumer / packager of this library the chance to opt-out of the vendored dependencies and use pre-compiled packages instead.

The changes here are made in such a way that using the vendored version of the libraries is the default, to preserve the current behavior.

Instead of exposing the whole `3rdparty` folder as a include dir, I've added the required CMakeLists files to make each vendored dependency build as a standalone library, which is cleaner and also provides clean bundries between the vendored and non-vendored dependencies.

There are some aspects I haven't verified, like the generated cmake config files. I'm not really sure how to test those.
There is also some more cleaning up required around the `ament_build.cmake` and `conan_build.cmake` modules, particularly the `export_btcpp_package` macro.


facontidavide:
thanks, but this completely breaks compilation in ROS, apparently. check CI

ericriff:
> thanks, but this completely breaks compilation in ROS, apparently. check CI

Yes, I've made it a draft to see how it goes with CI. I'll take a look.

cmake/conan.cmake
ericriff, line 1: I think this file can be deleted, it is not used.

ericriff:
@facontidavide CI is passing now. Let me know what you think.

3rdparty/flatbuffers/flatbuffers/base.h
ericriff, line 1: This file was moved into a `flatbuffers` folder so it can be included like this
```c++
#include <flatbuffers/base.h>`
```
Otherwise it would have to be included like this
```c++
#include <base.h>`
```
Which doesn't match the conan package.

CMakeLists.txt
```diff
@@ -19,6 +19,19 @@ option(USE_AFLPLUSPLUS "Use AFL++ instead of libFuzzer" OFF)
...
+option(USE_VENDORED_CPPSQLITE "Use the bundled version of cpp-sqlite" ON)
+option(USE_VENDORED_CPPZMQ "Use the bundled version of cppzmq" ON)
+option(USE_VENDORED_FLATBUFFERS "Use the bundled version of flatbuffers" ON)
+option(USE_VENDORED_LEXY "Use the bundled version of lexy" ON)
+option(USE_VENDORED_MINICORO "Use the bundled version of minicoro" ON)
+option(USE_VENDORED_MINITRACE "Use the bundled version of minitrace" ON)
+option(USE_VENDORED_TINYXML2 "Use the bundled version of tinyxml2" ON)
+option(USE_VENDORED_WILDCARDS "Use the bundled version of wildcards" ON)
+
+set(BTCPP_LIB_DESTINATION     lib)
```
ericriff, line 31: This code was repeated on both `conan_build.cmake` and `ament_build.cmake`. I've unified it here.

CMakeLists.txt
```diff
@@ -61,12 +74,6 @@ if(USE_V3_COMPATIBLE_NAMES)
     add_definitions(-DUSE_BTCPP3_OLD_NAMES)
 endif()
 
-#---- Find other packages ----
-find_package(Threads REQUIRED)
-
-
-set(BEHAVIOR_TREE_LIBRARY ${PROJECT_NAME})
```
ericriff, line 68: This was not used anywhere. There is a similar variable `set(BTCPP_LIBRARY ${PROJECT_NAME})`  which is used. I kept that one.

CMakeLists.txt
```diff
@@ -84,19 +91,67 @@ if ( ament_cmake_FOUND )
     include(cmake/ament_build.cmake)
 else()
     message(STATUS "------------------------------------------")
-    message(STATUS "BehaviorTree is being built with conan.")
+    message(STATUS "BehaviorTree is being built without AMENT.")
```
ericriff, line 94: I've changed this message to better reflect what's going on. This code path doesn't necesarily means that conan is being used, it just means that ament is not used. On the docs it says that one can build without conan or ament if ZeroMQ and Sqlite3 are provided somehow (e.g. apt on ubuntu). If you do that, then you fall on this code path and the log wrongly says that conan is being used.

CMakeLists.txt
```diff
@@ -179,8 +232,14 @@ target_link_libraries(${BTCPP_LIBRARY}
...
+        foonathan::lexy
+        minitrace::minitrace
+        tinyxml2::tinyxml2
+        minicoro::minicoro
+        flatbuffers::flatbuffers
+        cpp-sqlite::cpp-sqlite
+        cppzmq
+        wildcards::wildcards
     PUBLIC
         ${BTCPP_EXTRA_LIBRARIES}
```
ericriff, line 244: I still need to figure out this line. I think it is always empty now.

README.md
```diff
@@ -61,11 +61,11 @@ Three build systems are supported:
 
 Compiling with [conan](https://conan.io/):
 
-Assuming that you are in the **parent** directory of `BehaviorTree.CPP`:
+Assuming that you are in the **root** directory of `BehaviorTree.CPP`:
 
 ```
 mkdir build_release
-conan install . -of build_release -s build_type=Release
+conan install . -of build_release -s build_type=Release --build=missing
```
ericriff, line 68: `--build=missing` intructs conan to build whatever package is missing. For example I worked on this PR on Ubuntu 24.04 using gcc13 and conan-center had no binaries to serve me. The usage of this flag is trivial knoledge for someone who knows this way around conan, but it can be very helpul for someone who is taking his first steps with it.

cmake/FindZeroMQ.cmake
```diff
@@ -63,5 +63,13 @@ else (ZeroMQ_LIBRARIES AND ZeroMQ_INCLUDE_DIRS)
   # show the ZeroMQ_INCLUDE_DIRS and ZeroMQ_LIBRARIES variables only in the advanced view
   mark_as_advanced(ZeroMQ_INCLUDE_DIRS ZeroMQ_LIBRARIES)
 
+  if(ZeroMQ_FOUND AND NOT TARGET libzmq)
+    add_library(libzmq UNKNOWN IMPORTED)
```
ericriff, line 67: Using variables like `ZeroMQ_INCLUDE_DIRS` or `ZeroMQ_LIBRARIES` is old-school cmake and it is discouraged. All modern codebases should use targets instead. 
So I slightly tweaked this Find module to also expose a target, using a name that matches conan's.

cmake/conan_build.cmake
```diff
@@ -1,24 +1,19 @@
 list(APPEND CMAKE_PREFIX_PATH "${CMAKE_BINARY_DIR}")
 
 if(BTCPP_GROOT_INTERFACE)
-    find_package(ZeroMQ REQUIRED)
+#    find_package(ZeroMQ REQUIRED)
     list(APPEND BTCPP_EXTRA_LIBRARIES ${ZeroMQ_LIBRARIES})
     list(APPEND BTCPP_EXTRA_INCLUDE_DIRS ${ZeroMQ_INCLUDE_DIRS})
```
ericriff, line 6: I need to clean this up.

conanfile.py
ericriff, line 1: I've replaced the basic conanfile.txt with the more powerful python variant, mainly to be able to set some CMake variables on the toolchain to opt-out of the vendored libraries when conan provides them.
I left some of these variables commented out because conan doesn't support those packages yet. I also left a couple of messages for posterity: `sqlite3` and `zeromq` should not be listed here, since those are transitivce dependencies of the actual packages we use (`cppzmq` and `cpp-sqlite`), but those packages are not available con conan center.

sample_nodes/CMakeLists.txt
```diff
@@ -1,7 +1,4 @@
-include_directories( ../include )
```
ericriff, line 1: This line was not needed

src/loggers/zmq.hpp
ericriff, line 1: This file was duplicated, so I deleted this copy.
This header is the impl of `cppzmq`, and there is another copy on `3rdparty/cppzmq` (of course I kept that one).
Furthermore the versions didn't match. This one was 4.7 IIRC and on the 3rdparty folder we have 4.10. This could have caused some headaches.

tests/CMakeLists.txt
```diff
@@ -55,6 +55,6 @@ else()
 
 endif()
 
-target_link_libraries(behaviortree_cpp_test ${BTCPP_LIBRARY} bt_sample_nodes foonathan::lexy)
-target_include_directories(behaviortree_cpp_test PRIVATE include ${PROJECT_SOURCE_DIR}/3rdparty)
```
ericriff, line 59: Exposing the whole 3rdparty folder as a include dir is a bad pattern. We should only link in what we need

tools/CMakeLists.txt
```diff
@@ -1,17 +1,24 @@
...
 # install(TARGETS bt4_log_cat
 #         DESTINATION ${BTCPP_BIN_DESTINATION} )
 
-if( ZMQ_FOUND )
-    add_executable(bt4_recorder         bt_recorder.cpp )
-    target_link_libraries(bt4_recorder  ${BTCPP_LIBRARY} ${ZMQ_LIBRARIES})
-    install(TARGETS bt4_recorder
-            DESTINATION ${BTCPP_BIN_DESTINATION} )
-endif()
+# FIXME! This target doesn't build because behaviortree_cpp/flatbuffers/BT_logger_generated.h
```
ericriff, line 7: I left this big comment to bring attention to this issue

uilianries (commented):
Doing great, please, feel free to ask for help to continue this PR! Regards!

cmake/conan.cmake
ericriff: I edited it by mistake when I blanket-replaced all `conanfile.txt` references with `conanfile.py`, then I noticed it while reviewing the changes and left this comment.
I can revert these changes and remove it on a follow up or delete it now. I don't mind either way.

3rdparty/flatbuffers/CMakeLists.txt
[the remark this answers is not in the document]
ericriff: I created these CMakeLists.txt to make each vendored library be a CMake target. Then I matched the target name that conan exposes. 

Then you can use this pattern to select where `someLibrary` is coming from
```cmake
if(USE_VENDORED_someLibrary)
  add_subdirectory(3rdparty/someLibrary)
else()
  find_package(someLibrary REQUIRED)
endif()
```

But regardless of the origin of the package (vendored, package manager) you link it the same way
```cmake
target_link_library(btcpp PRIVATE someLibraryTargetName)
```

This also allowed me to remove an ugly line that was exposing the whole `3rdparty` folder as a include dir. it was something like this
```cmake
target_include_directory(btcpp PUBLIC 3rdparty)
```
This line was problematic because all the vendored headers where visible to btcpp, regardless of the value of `USE_VENDORED_someLibrary`. So if you opted out of some vendored lib and replaced it with a package, it was still not clear which version of the header was going to end up being included. The one from the package or the one vendored? It was a recipe for disaster.

Besides, CMake best practices is to always use targets.

conanfile.py
```diff
@@ -0,0 +1,37 @@
...
+from conan.tools.cmake import CMakeToolchain, CMakeDeps
+
+class BehaviortreeCppConan(ConanFile):
+    name = "behaviortree.cpp"
+    settings = "os", "arch", "compiler", "build_type"
+
+    default_options = {
+        "flatbuffers/*:header_only": True,
+    }
+
```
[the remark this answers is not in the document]
ericriff: How would you handle this then? The vendored version is headers only. I wanted to match that.

conanfile.py
```diff
@@ -0,0 +1,37 @@
...
+        self.requires("sqlite3/3.40.1") # This should be a transitive dependency of cpp-sqlite
+        self.requires("tinyxml2/10.0.0")
+        self.requires("zeromq/4.3.4") # This should be a transitive dependency of cppzmq
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+
+        #tc.variables["USE_VENDORED_CPPSQLITE"] = False
+        #tc.variables["USE_VENDORED_CPPZMQ"] = False
+        tc.variables["USE_VENDORED_FLATBUFFERS"] = False
```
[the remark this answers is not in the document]
ericriff: So if I use cache variables they don't end up on the toolchain file and then they never reach the build. Where are they set if not on the toolchain? on the presets? if that's the case then I should rework the docs and the CICD to use presets instead, and that requires a fairly modern CMake. Right? Not sure if that's worth it on this PR.
```markdown
Compiling with [conan](https://conan.io/):

Assuming that you are in the **root** directory of `BehaviorTree.CPP`:

```
mkdir build_release
conan install . -of build_release -s build_type=Release --build=missing
cmake -S . -B build_release -DCMAKE_TOOLCHAIN_FILE="build_release/build/Release/generators/conan_toolchain.cmake"
cmake --build build_release --parallel
```

```

facontidavide:
thanks a lot for the hard work. You don't know how much I appreciate that.

Do we need that level of granularity with vendor libraries? I am thinking that either I want all of them, or none of them, in practice...

what about a single option **USE_VENDORED_DEPENDENCIES**?

ericriff:
> Do we need that level of granularity with vendor libraries? I am thinking that either I want all of them, or none of them, in practice...
> 
> what about a single option **USE_VENDORED_DEPENDENCIES**?

We _need_ the granularity because conan doesn't support alll the libraries that btcpp vendors ATM.
I have a PR open to add support for `cpp-sqlite` https://github.com/conan-io/conan-center-index which will take us one step further. But some other dependencies like `Lexy` are not yet supported and there are no PRs open for it.
Then some other problematic dependencies are:
1. `cpp-zmq`: There is a note on the conan recipe that says that btcpp uses a custom version of this library. I don't know if that's true TBH -- https://github.com/conan-io/conan-center-index/blob/master/recipes/behaviortree.cpp/all/conanfile.py#L150
2. `wildcards`: It looks like this project has been abandoned 6 years ago, and they never released version 1.5.0 (even though the code is there and it is what btcpp is currently vendoring). So the latest supported version in conan is 1.4.0, which is not compatible with btcpp. We could address this by creating a custom version on conan, but that won't be enough either. I was planning on doing it anyways on a follow up.

So the short answer is yes, we need to have this level of granularity because we can't ensure that all package managers support all the dependencies this project has. If you check the conanfile, I left some lines commented out. Those are the libraries conan doesn't support yet.

facontidavide:
Ok, I can reimplements wildcard from scratch using AI.

About cpp-zmq, let me upgrade to the latest supported by conan

ericriff:
What do you think about doing that on a follow up? So we can limit the scope of this PR.

ericriff:
I've pushed a commit to use conan's cppzmq. Locally the build passed, so maybe this comment was wrong or outdated https://github.com/conan-io/conan-center-index/blob/master/recipes/behaviortree.cpp/all/conanfile.py#L150

uilianries:
About `cpp-sqlite`: I don't think it will be accepted in CCI for several reasons:

- The project is abandoned, 2 years since last commit
- It's not really popular, which means no maintenance guarantee
- No other package manager has it, not even Arch Linux or any other distro


> We need the granularity because conan doesn't support alll the libraries that btcpp vendors ATM.

Conan should not be the single source of truth. People are still free to use their own system, so an option like `USE_VENDORED_DEPENDENCIES` for all external dependencies sounds much easier. In case no being able to find externally, it still falls back on the vendorized versions.

ericriff:
>Conan should not be the single source of truth. 

Agreed, I'm mentioning conan because that's explicitely what it is being used here. But any package manager should be able to provide these dependencies. That's why I change a log message from "Building with Conan" to "building without ament".

> People are still free to use their own system, so an option like USE_VENDORED_DEPENDENCIES for all external dependencies sounds much easier. In case no being able to find externally, it still falls back on the vendorized versions.

I don't really like this approach. I've been bitten by it many times already.
Imagine we incude a single option, `USE_VENDORED_DEPENDENCIES`, which make CMake prefer the vendored stuff. If set to `FALSE` we do:

```cmake
find_package(libFoo)
if (NOT libFOO_FOUND)
  add_subdirectory(3rdparty/libFoo)
endif()
``` 
This has, in my eyes, a number of drawbacks.
1. Reproducible builds: I don't have libFoo installed and my builds work fine, because I'm using the vendored variant. The I inadvertedly install libFoo (lets say on an apt call, as a dependency of some other pacakge). Now btcpp builds start picking up that version, the builds break and I don't know why. Not everyone is a CMake expert to know where to look. Maybe I pick back the work on this repo 2 weeks after I did that and I forgot.
2. Works in my PC(™): The classic situation where it builds for me but not on a colleagues PC. This is because one of us is using libFoo from a package manager and the other is using the vendored version. We build with the same cmds, are on the same version of ubuntu and so on. But our build environments are slightly different.
3.  Conan target names: The libFoo recipe gets updated and the target name changed (maybe it was done explicitely, maybe as a side effect of changing the generator, maybe during the conan v1 -> v2 migration of a recipe). Now instead of `libFoo` conan provides a package named `libfoo`. Now you silently started using vendored instead of conan.

On my day job I maintain the build system of a massive repo where a bunch of engenieers contribute to, so I've been bitten by the things I listed above countless times. That's why I prefer rigid structures where unexpected scenarios produce clear errors and not subtle, hard to debug, behavior changes. And my _arch nemesis_ is the CMake find_package + if _FOUND pattern. Life would be better if CMake had decided to make REQUIRED the default on find_packge :joy:

With my proposal if you do
```cmake
set(USE_VENDORED_lifFoo ON)

if(NOT USE_VENDORED_lifFoo)
  find_package(libFoo REQUIRED)
else()
  add_subdirectory(3rdparty/libFoo)
endif()
```

The 3 problems listed above are greatly minimized. There still some variation when it comes to which package manager is serving you the package, but al least things are way more predictable. I've even seen patches on conan-center where the `REQUIRED` keyword is added to a find_package call. I suspect because of this.
But I understand how the other option is simpler. I just wanted to make its "hidden cost" as explicit as possible.

conanfile.py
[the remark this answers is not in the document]
uilianries: @ericriff Sorry, I was in a hurry when I tried to write. I just edited my message.

uilianries:
> I don't really like this approach. I've been bitten by it many times already.

@ericriff Okay, I buy it, as mixing dependencies from a package manager vs vendorized can pass without a warning. Thank you for detailing your real case!

facontidavide:
once again, if you feel that **cpp-sqlite** is a blocker, I can just reimplement something similar with 30 minutes of vibe-coding and remove this altogether from the repo... just saying

ericriff:
> once again, if you feel that **cpp-sqlite** is a blocker, I can just reimplement something similar with 30 minutes of vibe-coding and remove this altogether from the repo... just saying

That's up to you. Same with wildcards. I don't know much about this libraries, but they seem abandoned. It may be a good idea not to depend on them.

uilianries:
> once again, if you feel that **cpp-sqlite** is a blocker, I can just reimplement something similar with 30 minutes of vibe-coding and remove this altogether from the repo... just saying

It would be great! But in case you want to use some more maintained and active project for sqlite cpp wrapper, I would say [SQLiteCpp](https://github.com/SRombauts/SQLiteCpp). In its README, it pointed a bunch of other SQLite wrappers too, including some that are header-only (if you prefer)

conanfile.py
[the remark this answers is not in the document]
uilianries: > Unless we also enable the builenv?

Yes, will need to use build env. But adding a note about using system cmake 2.23 with conan is also not bad. 

In the future, Conan will have `conan run` command, which should make it more implicit. The feature is still under development, but you can take a look: https://github.com/conan-io/conan/pull/18972

ericriff:
So I was under the impression that Lexy was not supported on conan ATM. Mainly because of https://github.com/conan-io/conan-center-index/blob/master/recipes/behaviortree.cpp/all/conanfile.py#L106 and because I looked for it as `lexy`. But it is available, just with a different name (`foonathan-lexy`). So in the last commit I added support to consume that package from conan as well.

facontidavide:
@ericriff  FYI: https://github.com/BehaviorTree/BehaviorTree.CPP/commit/58ed0ece368395c43e5e88b4059345c8de444fb5

ericriff:
@facontidavide it looks like using cmake presets on the conan builds opened a can of worms.
Windows builds (I think) were using C++14 and now are actually using 17. This made a bunch of compile time errors pop up on `src/xml_parsing.cpp`. Silly things like using `or` instead of `||`. But I do see some warnings on the CI builds. Could you take a look? I don't have a windows box.

facontidavide:
https://github.com/BehaviorTree/BehaviorTree.CPP/commit/624b11b9c80f2af7ef7651cb02fb1341209debb8

I am going to create a new tagged release, 4.7.3

please refer to that one

ericriff:
Could you tag after this PR goes in? So we can create a new package in conan with all these packaging changes.

facontidavide:
sorry for pulling the rug under your PR with my changes, but I prefer to remove these dependencies now, before we handle with them in conan

https://github.com/BehaviorTree/BehaviorTree.CPP/releases/tag/4.7.3

facontidavide:
> Windows builds (I think) were using C++14 and now are actually using 17. This made a bunch of compile time errors pop up on src/xml_parsing.cpp. Silly things like using or instead of ||. But I do see some warnings on the CI builds. Could you take a look? I don't have a windows box.

I will look into this. thanks

ericriff:
This is looking good now!

facontidavide:
Silly question but how do I reproduce the issue you are talking about in **CI**?

I added naively `set(CMAKE_CXX_STANDARD 17)` but it seems to compile just fine

https://github.com/BehaviorTree/BehaviorTree.CPP/actions/runs/18145002973/job/51644676948?pr=1013

ericriff:
Revert these changes and the errors will pop up https://github.com/BehaviorTree/BehaviorTree.CPP/pull/1012/commits/314bdd7f010b7a4c8659c3e0ed7a351aadd953ab

They started showing up when I added https://github.com/BehaviorTree/BehaviorTree.CPP/pull/1012/commits/f88b72709d394511dbd68d5006bbd4ad2f9538ad

I had to do that because foonathan-lexy errors out if conan is set up to use c++14, and that's what `conan profile detect` was setting on windows. You can see it in the logs

Even with my fixed from above, I see some warnings on the logs. IIRC, when the tests run. They're not errors so CI passes.

facontidavide:
pushed  https://github.com/BehaviorTree/BehaviorTree.CPP/commit/d8c16098a49b7fd2718d1916c4d5df74173a859a

facontidavide:
hi, what is the status of this PR?

Ready to merge?

ericriff:
I think it is good to go. Builds and tests are passing.
We can address anything that pops up on a follow up PR.

ericriff:
@facontidavide would it be possible to release a new tag so I can create a conan package with these changes? They massively simplify packaging this lib

facontidavide:
sure thing, but let me first address the memory leak problem

ericriff:
There is a draft PR linked to the mem leak issue that adds ASan and TSan. You can use it to validate your fix.
We can't merge it untill the problems it flags are addressed. I can't work on it unfortunately.

uilianries:
Thank you all for working on this PR! :clap:

facontidavide:
@ericriff 4.8.0 is out

Issue #1014, opened 2025-10-05
Segfault due to incompatible tinyxml2 version


wentasah:
**Describe the bug**

BehaviorTree.CPP vendors tinyxml2 version 10.0.0. If a library user uses tinyxml2 too, but links to version 11.0.0, it can result in segfaults. This happens when ROS, and specifically nav2_behavior_tree is compiled against tinyxml2 11.0.0. See the backtrace below.

I can see two possible solutions:
1. Allow (optional) compilation of BehaviorTree.CPP against system-provided tinyxml2.
2. Export the tinyxml2 header and library used by BehaviorTree.CPP so that library users can link to the correct version.

IMO, solution 1 would be simpler.

Backtrace:
```
#0  0x00007ffff40a85be in __GI___libc_free (mem=0x1) at malloc.c:3375
#1  0x00007ffff64fdde7 in tinyxml2::DynArray<char, 20>::~DynArray (this=0x7fffffff03f0, __in_chrg=<optimized out>)
    at /build/behaviortree_cpp_v4-release-release-jazzy-behaviortree_cpp-4.7.1-1/3rdparty/tinyxml2/tinyxml2.h:215
#2  tinyxml2::XMLPrinter::~XMLPrinter (this=0x7fffffff02e0, __in_chrg=<optimized out>) at /build/behaviortree_cpp_v4-release-release-jazzy-behaviortree_cpp-4.7.1-1/3rdparty/tinyxml2/tinyxml2.h:2251
#3  0x00007ffff6431bd1 in BT::XMLParser::PImpl::loadDocImpl (this=<optimized out>, doc=<optimized out>, add_includes=add_includes@entry=true)
    at /nix/store/2agih0y5ns3sgbviw2xhivdgg59b41g9-gcc-14-20241116/include/c++/14-20241116/bits/allocator.h:182
#4  0x00007ffff64fbda8 in BT::XMLParser::loadFromFile (this=<optimized out>, filepath=..., add_includes=true)
    at /build/behaviortree_cpp_v4-release-release-jazzy-behaviortree_cpp-4.7.1-1/src/xml_parsing.cpp:164
#5  0x00007ffff6473309 in BT::BehaviorTreeFactory::createTreeFromFile (this=0xbcaf28, file_path=..., blackboard=...)
    at /build/behaviortree_cpp_v4-release-release-jazzy-behaviortree_cpp-4.7.1-1/src/bt_factory.cpp:366
[...]

PR #1015, opened 2025-10-05
CMake: prefer system tinyxml2

nim65s:
Hi,

Here is a simple fix for #1014.


facontidavide:
Broken CI

facontidavide:
please check this commit: 
https://github.com/BehaviorTree/BehaviorTree.CPP/commit/c381486027294aa95f5d5a503810b031ebdd6f84

nim65s:
That looks great, thanks !

facontidavide:
Do we want to set USE_VENDORED_TINYXML2 by default when compiling in ROS2?

nim65s:
From a downstream distribution point of view, I would say generally vendoring should be avoided to ensure the cohesion of the entire software distribution, ease upgrading, avoid security issues, and consume less resources.

facontidavide:
developing a library for multiple operative systems, build systems (colcon, cmake, piki, vanilla cmake, etc) is challenging.

Yes, I am guilty of vendoring libraries... 

For the records, I tried hard to make these private and not public dependencies, that is extra work too.

nim65s:
I know the challenges and I blame you for nothing. Quite the contrary, I'm just here to thank you for the software, and to provide assistance on the packaging side, if needed.

(This is actually my work, and I love it 😅)

Making the dependencies private is a great move that few project do, thanks for that too !

PR #1016, opened 2025-10-06
Remove unused conan.cmake

ericriff:
The deletion of this file was discussed on https://github.com/BehaviorTree/BehaviorTree.CPP/pull/1012#discussion_r2384490369


facontidavide:
great

PR #1020, opened 2025-10-09
Include `curses.h` instead of `ncurses.h`

lalten:
`ncurses.h` is usually a symlink to the actual header, `curses.h`. See also https://stackoverflow.com/questions/59555474/difference-between-the-headers-ncurses-h-and-curses-h

When building with Bazel (thanks @kgreenek for adding in https://github.com/bazelbuild/bazel-central-registry/pull/5154), you (currently) want to depend on `curses.h`, otherwise you may end up depending on a host installed version accidentally


kgreenek:
Thank you for this! I have had to require that libncurses-dev is installed on the system and did not understand why.

PR #1024, opened 2025-10-17
Allow build with BTCPP_SHARED_LIBS=OFF flag

adiego73:
- Add install targets to 3rdparty dependencies
- Add cmake version config


facontidavide:
no, these are and must be private dependencies that are NEVER installed in the system.

The solutions should be anbother one

facontidavide:
Do you mind testing #1027 ?

adiego73:
> no, these are and must be private dependencies that are NEVER installed in the system.
> 
> The solutions should be anbother one

Makes sense. I will test #1027 as soon as possible.

adiego73:
Looks good to me. However, lexy is still being installed in my system; I don't know if this is intended or not.

facontidavide:
Not intended, I will try fixing it

facontidavide:
pushed the fix. Lexy should not be installed anymore

PR #1026, opened 2025-10-22
Properly mark sqlite3 and cppzmq as public dependencies

ericriff:
I made them `PRIVATE` as a side effect of https://github.com/BehaviorTree/BehaviorTree.CPP/pull/1012 
This introduces linking issues on some scenarios (if bt and its dependencies are built as shared libs, then the targets that depend on bt like the tests, tools and examples fail to link with undef symbols).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,9 +231,9 @@ target_link_libraries(${BTCPP_LIBRARY}
         tinyxml2::tinyxml2
         minicoro::minicoro
         flatbuffers::flatbuffers
+    PUBLIC
         $<$<BOOL:${BTCPP_GROOT_INTERFACE}>:cppzmq>
         $<$<BOOL:${BTCPP_SQLITE_LOGGING}>:SQLite::SQLite3>
-    PUBLIC
         ${BTCPP_EXTRA_LIBRARIES}
 )
 


facontidavide:
I am not a cmake expert but isn't the right solution instead to ass **$<BUILD_INTERFACE:** ?

ericriff:
I've only seen  $<BUILD_INTERFACE: in the context of include directories. That would help with not found headers.
This is for linking issues.
And TBH I didn't do a super deep dive here, I noticed the linking issues and then compared current master with master before my changes from https://github.com/BehaviorTree/BehaviorTree.CPP/pull/1012. That made me realize that on my PR I moved these two deps from `PUBLIC` to `PRIVATE` (they used to be part of `BTCPP_EXTRA_LIBRARIES`). So I moved them back to PUBLIC and that took care of it.

facontidavide:
The change in this PR **did** fix the problem when compiling with 
```
cmake -S . -B build  -DBTCPP_SHARED_LIBS=OFF
```

But more need to be done in the context of conan, apparently

ericriff:
`BTCPP_SHARED_LIBS` only affects libraries built on this project. When you pull pre-compiled libraries (from conan or any other package manager, even a system pkg manager) I believe it has no effect.

facontidavide:
yes, I know. This are two separate issues

PR #1030, opened 2025-10-23
Event-based tree ticking

pleemann:
I have a tree, which checks some pre-conditions in our system state, using a ReactiveSequence, and executes some (re)actions if they are not met. The pre-conditions are checked using a simple custom node, which reads the system state.
I have a callback mechanism, which notifies me when the system state changes, and I would like to tick the tree based on these events instead of a fixed rate. The callbacks can be executed from different threads, so I would like to avoid ticking the tree in the callback directly.

What seems to work is something like this:

```
void ReactionManager::workerMain() {
  while (!exitWorkerLoop_) {
    tree_->tickOnce();
    tree_->sleep(std::chrono::milliseconds(10000));
  }
}

void ReactionManager::wakeupWorker() {
  if (tree_) {
    if (!tree_->subtrees.empty()) {
      auto& nodes = tree_->subtrees.front()->nodes;
      if (!nodes.empty()) {
        nodes.front()->emitWakeUpSignal();
      }
    }
  }
}
```
However, I wondered why the `Tree` itself does not have such a `emitWakeUpSignal`? This pull request adds this method.

### DIFF
--- a/include/behaviortree_cpp/bt_factory.h
+++ b/include/behaviortree_cpp/bt_factory.h
@@ -119,13 +119,19 @@ public:
   [[nodiscard]] TreeNode* rootNode() const;
 
   /**
-    * @brief Sleep for a certain amount of time. This sleep could be interrupted by the method TreeNode::emitWakeUpSignal()
+    * @brief Sleep for a certain amount of time. This sleep could be interrupted by the methods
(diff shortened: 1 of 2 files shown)


facontidavide:
makes sense

PR #1035, opened 2025-10-30
Remove tinyxml2_vendor

mini-1235:
Remove tinyxml2_vendor, this is not needed, see https://github.com/ros2/ros2/issues/1729 and https://github.com/ros2/ros2/pull/1751


mini-1235:
Ah, I forgot the main branch is also for humble, I will close this instead

PR #1060, opened 2026-01-19
Remove the `any` contrib library

nyibbang:
## Summary

This change completely removes the `any` contrib library and replaces its use with `std::any` available in C++17.

## Motivation

This change has several benefits:
- It removes code from the project and uses the standard version instead.
- It removes a warning about a deprecated use of `std::aligned_storage` when compiling a project using the library in C++23 or more.

## Impact on API

This change removes a header and replaces uses of `linb::any` with `std::any` in `safe_any.hpp`. In theory, if `contrib/any.hpp` is considered an implementation details of `BT::Any` then this should not be considered an API break as the interface of the class does not change.

## Release Notes

* **Refactor**
  * Updated internal dependencies to use standard library components, improving code maintainability and reducing custom implementation overhead while maintaining full backward compatibility.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


facontidavide:
Thanks but this will probably affect the ABI and I will not do it.

Additionally, the implementation I use has a guaranteed 16 bytes small object optimization and I rely on that 

image

nyibbang:
Fair enough, would you be open to a PR to remove the use of aligned_storage in that case ?

PR #1075, opened 2026-02-01
Fix #948: parseString errors with custom enum types

### LINKED ISSUE #948: ParseString errors
**Describe the bug**
There seem to be fundamental errors in the implementation of parseString inside tree_node.h. Particularly, the find operation is unsafe on [L435](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/14589e589c5049fbb7f1fb30ccaa3efc66c84f18/include/behaviortree_cpp/tree_node.h#L435) if getInput is called during construction. This is because the unordered_map is not instantiated yet. However this is only a side-issue of a more fundamental oversight. 

It seems that it is not possible to properly declare your own enum and then parse strings into this, while ignoring the ScriptingEnums functionality. It seems this is a known issue to some extent as there is an exception made for NodeStatus on [L433](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/14589e589c5049fbb7f1fb30ccaa3efc66c84f18/include/behaviortree_cpp/tree_node.h#L433). The direct implication as it stands is that if someone were to use an InputPort for say NodeType, and safely getInput after construction, let's say in tick(), the parsing would be incorrect as it would [turn the string provided into an int and then cast that int with NodeType](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/14589e589c5049fbb7f1fb30ccaa3efc66c84f18/include/behaviortree_cpp/tree_node.h#L444C7-L444C58).

*How to Reproduce**

I made [a fork of btcpp_sample to exemplify the core of the problem](https://github.com/EGAlberts/btcpp_sample). 
Here is the output of running the main.cpp there while using the current commit of this repo:
```
./build/btcpp_sample/btcpp_sample
Inside tick() with getInput:
IDLE
Undefined

Without getInput: 
IDLE
Control
```

As you can see due to the exemption in ParseString for NodeStatus it comes out correctly, but for NodeType it doesn't.

You can also find commented out code to exemplify the unsafe operation in the constructor.

facontidavide:
## Summary
- Fix null pointer dereference on config().enums when parseString is called before enums are registered
- Fix enum types with convertFromString<T> specializations not being called
- Three-level fallback: scripting enums registry -> numeric via from_chars -> convertFromString<T>

## Test plan
- [x] New test: Enums.ParseEnumWithConvertFromString_Issue948
- [x] All existing tests pass (298/298)

Closes #948

### DIFF
--- a/include/behaviortree_cpp/tree_node.h
+++ b/include/behaviortree_cpp/tree_node.h
@@ -20,6 +20,7 @@
 #include "behaviortree_cpp/utils/strcat.hpp"
 #include "behaviortree_cpp/utils/wakeup_signal.hpp"
 
+#include <charconv>
 #include <exception>
 #include <map>
 #include <utility>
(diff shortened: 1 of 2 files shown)


RithwikU:
### When will this PR #1075 (enum parsing fix) be included in ROS apt packages?

Hi @facontidavide,

Thank you for merging PR #1075 which fixes the enum parsing issues! This addresses exactly the problem we're encountering.

#### Our situation:

- We're using ROS 2 Jazzy with ros-jazzy-behaviortree-cpp from apt
- Current installed version: 4.8.3-1noble.20260126.192859 (built Jan 26, 2026)
- Issue: Enum types registered with registerScriptingEnums cannot be parsed from XML when assigned via the := operator (e.g., throw_side:=LEFT fails with type mismatch errors)
- PR #1075 was merged Feb 1, 2026, which is after the 4.8.3 build

#### Question: When can we expect PR #1075 to be included in the ros-jazzy-behaviortree-cpp apt packages?

Will it be in:
- A 4.8.x patch release?
- A 4.9.0 release?
- Should we build from source in the meantime?

Understanding the timeline will help us decide whether to implement a workaround (custom convertFromString specialization) or wait for the apt package update.

Thanks for maintaining this excellent library!

PR #1110, opened 2026-02-04
Use `convertFromString<double>` instead of `std::from_chars`

chhtz:
This https://github.com/BehaviorTree/BehaviorTree.CPP/pull/1099 very recently broke compilation on older g++ versions (which don't properly support `std::from_chars` for floating point types).

@facontidavide

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,7 +51,7 @@ repos:
         entry: ./run_clang_tidy_hook.sh
         language: script
         files: \.(cpp|hpp|h)$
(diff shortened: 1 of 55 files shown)


chhtz:
No idea what caused the failing tests (to me it appears unrelated to this change).
Also, I'm fine if you want to drop support for old g++ versions (< g++-11). We would then pin this library to an old commit on our older systems.

facontidavide:
yeah, the main branch is failing too. do not worry. thanks for sending this

Issue #995, opened 2025-07-30
When Using Mock nodes allow conditional Success or Failure for return_status


pepeRossRobotics:
**Issue**
We have been implementing mock testing using the process described in [here](https://www.behaviortree.dev/docs/tutorial-advanced/tutorial_15_replace_rules) and we find that sometimes we want the mock to return SUCCESS or FAILURE based on a particular condition.  Currently we use the post_script to set up a flag and then in the tree we set a script that reacts to the flag. 

This solution is not ideal because adds a lot of unnecessary code to the tree to be able to test complex trees.

**Describe the solution you'd like**
It would be good to be able to set the return_status based on a script output e.g.

```
return_status = "(robot_failure == true) ? FAILURE : SUCCESS"
```

PR #1125, opened 2026-04-02
Add scripted return status support for TestNode mocks

pepeRossRobotics:
This adds support for computing a TestNode return status from a script, so mock behavior can depend on blackboard state without adding extra nodes to the tree.

It also updates the substitution tests and tutorial examples to cover both the observable and strict-failure mock flows.

Addresses https://github.com/BehaviorTree/BehaviorTree.CPP/issues/995


facontidavide:
Thanks for this contribution — the feature itself is welcome and addresses #995 well, the test coverage is good, and I verified locally that everything builds and passes.

However, there are a few things that need to be addressed before this can be merged:

**1. We can not accept API changes: `return_status` must remain a plain `NodeStatus`.**

Changing `TestNodeConfig::return_status` from `NodeStatus` to `std::optional<NodeStatus>` is a source-compatibility break for any downstream code that reads the field directly (e.g. `NodeStatus s = cfg.return_status;`, `toStr(cfg.return_status)`, or a `switch` on it).

The optional isn't actually needed for the feature: since `return_status_script` takes precedence over the fixed status anyway (as your own documentation states), you can keep

```cpp
NodeStatus return_status = NodeStatus::SUCCESS;
```

unchanged. Concretely:
- drop the `.reset()` calls from the examples and tests;
- in `loadSubstitutionRuleFromJSON`, use a local "was `return_status` present in the JSON" flag for the *"must contain return_status or return_status_script"* validation, instead of resetting the member.

That makes the whole PR purely additive.

**2. Misleading error message for out-of-range script results.**

`validateTestNodeStatus()` hardcodes *"resolved to invalid status IDLE"*. A script that evaluates to e.g. `7` reports "IDLE", which would send users debugging in the wrong direction. Please include the actual offending value in the message (the raw integer is fine, since `toStr` can't render invalid enum values).

**3. Unrelated change in `tests/gtest_wakeup.cpp`.**

The `system_clock` → `steady_clock` switch and the relaxed assertion bound are reasonable on their own, but they are unrelated to this feature and not mentioned in the PR description. Please move them to a separate PR (or at minimum call them out explicitly in the description).

**4. Please leave the existing tutorial `t15_nodes_mocking.cpp` untouched.**

Tutorial 15 corresponds to a published page on behaviortree.dev; restructuring its tree (the added `Fallback`, the mock flag) makes the code diverge from the website text. Keep the original example as it is and demonstrate the scripted return status only in your new `t15_nodes_mocking_strict_failure.cpp` (or a dedicated new example).

pepeRossRobotics:
Comments addressed, also opened https://github.com/BehaviorTree/BehaviorTree.CPP/pull/1164

facontidavide:
@copilot resolve the merge conflicts in this pull request

facontidavide:
@copilot resolve the merge conflicts in this pull request

PR #1127, opened 2026-04-04
Add `Tree::wakeUpSignal()` getter

tonynajjar:
## Problem

`Tree::sleep()` uses `WakeUpSignal::waitFor()`, which internally calls `std::condition_variable::wait_for()` — this always measures time against `steady_clock` (wall time). When using a simulated clock that runs faster or slower than real time (e.g., ROS 2's `/clock` topic with `use_sim_time`), the tree tick rate doesn't scale with the simulation. At 10x sim speed, a 100ms loop timeout still blocks 100ms real time, so the tree ticks 10x too slowly relative to the simulated world.

C++ provides no mechanism to make `condition_variable` wait on a custom clock, so any solution requires polling.

## Solution

Add `Tree::wakeUpSignal()` — a simple getter that returns the tree's shared `WakeUpSignal`. This allows frameworks to implement their own clock-aware sleep logic externally while still being able to check for wake-up signals (preemption).

For example, in ROS 2 (Nav2), the between-tick sleep loop can poll the ROS clock for the deadline while doing short `waitFor(1ms)` calls to stay responsive to preemption:

```cpp
auto wake_up = tree->wakeUpSignal();
while (ros_clock->now() < deadline) {
  if (wake_up->waitFor(std::chrono::milliseconds(1)))
    return true;  // woken up early
}
```

### DIFF
--- a/include/behaviortree_cpp/bt_factory.h
+++ b/include/behaviortree_cpp/bt_factory.h
@@ -127,7 +127,7 @@ public:
   /// WakeUpSignal so it can check for early wake-up.
   /// Return true if woken up before the timeout, false if the full duration elapsed.
   using SleepOverrideFunc =
-    std::function<bool(std::chrono::system_clock::duration, WakeUpSignal&)>;
+      std::function<bool(std::chrono::system_clock::duration, WakeUpSignal&)>;
 
   /**
     * @brief Sleep for a certain amount of time. This sleep could be interrupted by the methods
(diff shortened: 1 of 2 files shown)


facontidavide:
thanks

SteveMacenski:
@facontidavide can we get a new release? This is blocking potential releases in Lyrical for Nav2

PR #1137, opened 2026-05-05
Fix LoopNode does not reload dynamic vector input on second execution

Rayman:
I've implemented the suggestion from https://github.com/BehaviorTree/BehaviorTree.CPP/issues/1124.

This seems to have fixed the issue I had with LoopNode not reloading the input vector on second execution. I've also added a unittest to prevent this issue in the future.


Rayman:
@facontidavide can you take a look at this PR?

PR #1148, opened 2026-06-03
Jazzy release 3.8.8

marcoag:
@facontidavide let's make a release of the `Jazzy` fix so we can remove the current [regression](https://repo.ros2.org/status_page/ros_jazzy_default.html?q=REGRESSION).


marcoag:
@facontidavide forgot to mention we still need a **tag** `3.8.8` a the top of the [v3.8 branch](https://github.com/BehaviorTree/BehaviorTree.CPP/tree/v3.8), please help with that. Once we have that I can take care and release it.

PR #1152, opened 2026-06-14
Use GNUInstallDirs with CACHE variables for install destinations (fixes #1120)

### LINKED ISSUE #1120: Allow overriding installation directories to support multiarch packaging (Debian/Ubuntu) and standalone usage
Hi there, thanks for this amazing library!

I am currently working on packaging `BehaviorTree.CPP` natively for Debian/Ubuntu systems using standard Debian packaging tools (`dpkg-buildpackage`). 

I understand that the installation paths in `CMakeLists.txt` might be hardcoded to ensure it builds seamlessly and consistently within the ROS/ROS2 ecosystem (e.g., ament/catkin). However, for users and system maintainers who want to use or package `BehaviorTree.CPP` as a standalone, general-purpose C++ library, this strict hardcoding causes some limitations.

### Description

In Debian-based systems, libraries should be installed into multiarch directories like `/usr/lib/x86_64-linux-gnu` instead of `/usr/lib/`. 
When configuring with CMake, I pass the argument `-DBTCPP_LIB_DESTINATION=lib/x86_64-linux-gnu`, but it gets overwritten during the build process because the paths are defined using normal `set()` commands without the `CACHE` option.

In `CMakeLists.txt`:
```cmake
set(BTCPP_LIB_DESTINATION     lib)
set(BTCPP_INCLUDE_DESTINATION include)
set(BTCPP_BIN_DESTINATION     bin)
```
### Expected behavior

It would be incredibly helpful if these destination paths could be overridden via CMake command-line arguments (e.g., -DBTCPP_LIB_DESTINATION=...) for non-ROS environments.

### Suggested Solution

The most standard CMake approach would be to use the GNUInstallDirs module, which automatically handles platform-specific installation paths.

Alternatively, simply making them CACHE variables would solve the issue without breaking any existing ROS workflows:

wooksong:
Fixes #1120

## Problem

The install destination variables use plain `set()`, which prevents overriding them from the CMake command line. This blocks Debian/Ubuntu packagers from installing to multiarch directories (e.g. `lib/x86_64-linux-gnu`).

## Solution

`include(GNUInstallDirs)` with `set(... CACHE STRING ...)` defaults.
 
Two override mechanisms:

 - `-DBTCPP_LIB_DESTINATION=lib/x86_64-linux-gnu` (project-specific)
 - `-DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu` (system-wide, standard)

## API / ABI / Backward compatibility

 No API or ABI changes: this is a build system change only.

 - **ROS2/ament builds**: unaffected. `ament_build.cmake` only calls `mark_as_advanced()` on these variables; ament handles installation separately.
 - **Standard Linux**: `CMAKE_INSTALL_LIBDIR` resolves to `lib`: no change from current behavior.
 - **Existing `-DBTCPP_LIB_DESTINATION=...` usage**: still works, CACHE variable takes precedence.

## Pre-commit / clang-tidy

Not applicable: this PR changes only `CMakeLists.txt`. The `clang-format` and `clang-tidy` hooks target `.cpp/.hpp/.h` files only.

## Verified

| Check | Result |
|-------|--------|
| Configure (default) | `BTCPP_LIB_DESTINATION: lib` |
| Build | 100% (45 TUs, no warnings) |
| Install (default) | `lib/libbehaviortree_cpp.so` |
| Override | `-DBTCPP_LIB_DESTINATION=custom/lib` -> `custom/lib/libbehaviortree_cpp.so` |

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,9 +28,11 @@ option(USE_VENDORED_FLATBUFFERS "Use the bundled version of flatbuffers" ON)
 option(USE_VENDORED_MINICORO "Use the bundled version of minicoro" ON)
 option(USE_VENDORED_MINITRACE "Use the bundled version of minitrace" ON)
 
-set(BTCPP_LIB_DESTINATION     lib)
-set(BTCPP_INCLUDE_DESTINATION include)
-set(BTCPP_BIN_DESTINATION     bin)
+include(GNUInstallDirs)
+
+set(BTCPP_LIB_DESTINATION     "${CMAKE_INSTALL_LIBDIR}"     CACHE STRING "Library installation directory")
+set(BTCPP_INCLUDE_DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}" CACHE STRING "Header installation directory")
+set(BTCPP_BIN_DESTINATION     "${CMAKE_INSTALL_BINDIR}"     CACHE STRING "Binary installation directory")
 
 set(BASE_FLAGS "")
 


SteveMacenski:
@facontidavide this breaks the ROS build farm binaries resulting in this conversation https://github.com/ament/ament_cmake/issues/630. 

[I noted here](https://github.com/ros-navigation/navigation2/issues/6123#issuecomment-5028537777) my workaround to temporarily use the binaries to get going 

I would recommend reverting since we cannot set those flags in the ros build farm & this blocks the release of packages that depend on BT.CPP, including Nav2.

facontidavide:
@SteveMacenski if my expensive slop generator isn't wrong, PR #1176  should fix the issue.

I will release 4.10.0

PR #1163, opened 2026-07-13
Fix package share path API for newer ament_index_cpp

mini-1235:
This updates the XML include handling to use the newer package share path API when available, while preserving the older fallback for older ament_index_cpp versions.

Related: https://github.com/ament/ament_index/pull/120

### DIFF
--- a/src/xml_parsing.cpp
+++ b/src/xml_parsing.cpp
@@ -50,7 +50,12 @@
 #include <map>
 
 #ifdef USING_ROS2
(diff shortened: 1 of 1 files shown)


mini-1235:
Without this PR, I am unable to compile BTCPP on ros:rolling

mini-1235:
@facontidavide I would appreciate if you could review and merge this PR so that we could avoid using my fork version in Nav2

mini-1235:
Thank you!

PR #1172, opened 2026-07-17
fix use-after-free from dangling StringView in Blackboard::getKeys

aysha-afrah26:
getKeys() hands back a std::vector<StringView> whose elements point into the std::string keys of the internal storage_ map. The shared lock added by the earlier BUG-6 fix is only held for the duration of getKeys() itself, so the returned views outlive it. ExportBlackboardToJSON walks those views after the lock is gone, and that path runs on the Groot2 publisher server thread in response to a BLACKBOARD request coming off the network. If the tree ticks an UnsetBlackboard node (or anything calling Blackboard::unset) in that window, unordered_map::erase destroys the key string and the StringView the exporter still holds now points at freed memory; constructing a std::string from it is a heap-use-after-free. I hit it while stress-testing concurrent key reads against unset, and ASan pins the read to the std::string copy in ExportBlackboardToJSON with the free coming from unset. The test file already had a note that getKeys() could dangle on erase, so this finishes that fix by returning owning std::string copies made under the lock instead of views into storage. Worth calling out that the return type changes from vector<StringView> to vector<std::string>, a small API change; all in-tree callers already copied into std::string anyway. The added regression test reads keys while another thread erases and reports the use-after-free under ASan before the change, clean after.

### DIFF
--- a/include/behaviortree_cpp/blackboard.h
+++ b/include/behaviortree_cpp/blackboard.h
@@ -123,7 +123,7 @@ public:
 
   void debugMessage() const;
 
-  [[nodiscard]] std::vector<StringView> getKeys() const;
+  [[nodiscard]] std::vector<std::string> getKeys() const;
 
   [[deprecated("This command is unsafe. Consider using Backup/Restore instead")]] void
   clear();
--- a/src/blackboard.cpp
+++ b/src/blackboard.cpp
@@ -137,16 +137,14 @@ void Blackboard::debugMessage() const
   }
 }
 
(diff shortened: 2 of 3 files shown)


aysha-afrah26:
any update?

facontidavide:
you are changing the API.

everything that use a string_view should make no assumption about the lifetime of the view.

So, even if your change is technically correct, I can not merge it.
The recommendation for the user is to assume that the output of getkeys is not persistent.

PR #1174, opened 2026-07-19
fix unbounded recursion in the script expression parser

aysha-afrah26:
The scripting engine compiles precondition, postcondition, and `<Script>` attributes into an AST with a hand-written recursive-descent Pratt parser, where every open parenthesis, unary prefix (`-`, `!`, `~`), and right-hand operand adds one `parseExpr` frame. Nothing bounds that recursion, so a script value that is a long run of `(` or `!` recurses once per character until the native stack is exhausted and the process dies with a SIGSEGV. Those strings come straight from XML attributes and are compiled at tree build and validate time, before any node ticks, so a single crafted tree crashes the executor while it is still loading (a ~30 KB nested script is enough on a default 8 MB stack, and less on Windows). I noticed it because the XML parser already guards against this exact shape with `kMaxNestingDepth` in `VerifyXML` and `recursivelyCreateSubtree`, while the script parser those paths feed had no equivalent limit. The fix caps the nesting depth inside `parseExpr` and throws a regular parse error once the limit is passed, matching the XML cap. Ordinary scripts are unaffected: long flat expressions and chained comparisons are parsed iteratively, so only genuinely deep nesting trips the limit, and the added test covers both the rejection and that normal scripts still parse.

### DIFF
--- a/src/script_parser.cpp
+++ b/src/script_parser.cpp
@@ -46,6 +46,14 @@ public:
 private:
   std::vector<Token> tokens_;
   size_t current_ = 0;
+  int depth_ = 0;
+
+  // Upper bound on the nesting of the recursive-descent grammar. Each open
+  // paren / unary prefix / right-hand operand adds one parseExpr frame, so a
+  // crafted expression like "((((...0...))))" would otherwise recurse until the
+  // native stack is exhausted. This mirrors the kMaxNestingDepth cap the XML
+  // parser already applies in VerifyXML and recursivelyCreateSubtree.
+  static constexpr int kMaxNestingDepth = 256;
 
   // Binding power constants.  Higher value = tighter binding.
   static constexpr int kAssignmentBP = 2;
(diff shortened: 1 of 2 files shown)


aysha-afrah26:
any thoughts?

fallenmi (approved):
This review was prepared with OpenAI Codex assistance under the account owner's authorization.

Codex independently verified exact head `5028cd8e152a95c6bd80c754b03617fab46e46c1` and its clean merge with current `master`. The depth check covers every recursive `parseExpr()` entry, successful parses restore the counter, and the public parsing APIs already convert the resulting `RuntimeError` into a normal validation error.

With a 1 MiB stack, an independent oracle reproduced the bug on the recorded base: depth 256 was still accepted and a 20,000-parenthesis expression exited with `SIGSEGV` (139). At both the exact head and current merge, depth 255 remains valid, depth 256 is rejected, a 1,000-term flat expression remains valid, and the 20,000-depth case returns a normal validation failure. The focused regression passed on head and merge, the current-merge parser slice passed 20/20 tests, `git diff --check` is clean, and all published CI statuses are green, including sanitizers, Windows Debug/Release, formatting, and clang-tidy.

No blocking findings.

PR #1178, opened 2026-07-22
reject trailing characters in SwitchNode numeric comparison

aysha-afrah26:
SwitchNode picks a child by comparing its `variable` against each `case_N` with CheckStringEquality, which falls back from a string compare to an integer compare and then a real compare. The `variable` is a blackboard entry, so in practice it often carries a value that came from outside the tree. The integer and real fallbacks call std::from_chars (and std::stoi/std::stod on toolchains without the floating-point from_chars) but never check that the whole string was consumed, so `5abc`, `42xxxx`, and even `5 ` parse as their leading number and compare equal to case `5`, `42`, and `5`. That routes a selector with trailing bytes to a case it does not actually equal, which is easy to slip past a check that only looked at the string form. I ran into it on a switch whose selector came from a message field. The fix requires full consumption (ptr == end, or pos == size on the stoi/stod path) at all four numeric parses; I also let the real comparison resolve scripting enums the way the integer one already does, so an enum name still matches the double a script stores for it. String equality and legitimate numeric matches like `5` versus `5.0` are unchanged.

### DIFF
--- a/src/controls/switch_node.cpp
+++ b/src/controls/switch_node.cpp
@@ -39,13 +39,15 @@ bool CheckStringEquality(const std::string& v1, const std::string& v2,
       }
     }
 #if __cpp_lib_to_chars >= 201611L
-    auto [ptr, ec] = std::from_chars(str.data(), str.data() + str.size(), result);
-    return (ec == std::errc());
+    const char* end = str.data() + str.size();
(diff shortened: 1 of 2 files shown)


facontidavide:
thanks

PR #1180, opened 2026-07-24
fix use-after-free of the timer node mutex on destruction

aysha-afrah26:
The Sleep, Delay, and Timeout nodes each own a TimerQueue whose worker thread runs the timer callback, and that callback locks a std::mutex that belongs to the node (delay_mutex_ or timeout_mutex_). Because the TimerQueue was the first-declared member it is destroyed last, after that mutex, so when a tree is torn down while a timer is still pending the worker thread (joined only in ~TimerQueue) can lock a mutex whose destructor has already run. On macOS/libc++ the pending handler then aborts the process with "mutex lock failed: Invalid argument"; on other standard libraries it is undefined behavior. It shows up as intermittent aborts in the timer and parallel teardown tests, including Parallel.Issue819_SequenceVsReactiveSequence, which the macOS CI currently skips as flaky. Declaring the TimerQueue as the last member fixes it, since ~TimerQueue() joins the worker thread while the mutex it locks is still alive. Runtime behavior for valid trees is unchanged; only the member declaration order, and therefore the destruction order, moves.

### DIFF
--- a/include/behaviortree_cpp/actions/sleep_node.h
+++ b/include/behaviortree_cpp/actions/sleep_node.h
@@ -40,11 +40,13 @@ public:
   }
 
 private:
-  TimerQueue<> timer_;
   uint64_t timer_id_ = 0;
 
   std::atomic_bool timer_waiting_ = false;
   std::mutex delay_mutex_;
+  // Keep last: ~TimerQueue() joins the worker thread, so this must be destroyed
+  // before delay_mutex_ and timer_waiting_, which the timer handler touches.
+  TimerQueue<> timer_;
 };
 
 }  // namespace BT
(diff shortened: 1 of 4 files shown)


aysha-afrah26:
any update?

facontidavide:
you are changing the ABI, but not providing a unit test to demonstrate the need for this change

aysha-afrah26:
Fair point. I pushed a regression test (Decorator.DestroyTreeWhileTimerIsPending): it creates a tree where Sleep, Delay and Timeout all have pending timers and destroys it in a loop. Without the header change it reliably aborts on libc++/macOS with "mutex lock failed: Invalid argument", since the canceled handler locks the node mutex after its destructor has run. With the change it passes, and the rest of the suite stays green.

On the ABI: you're right that the reorder moves the offsets of the private members (object size is unchanged). If you'd rather keep the layout stable, the alternative is to leave the declaration order alone and add a stop() method to TimerQueue that the node destructors call, so the worker thread is joined before any member is destroyed. Happy to rework it that way if you prefer.

facontidavide:
thanks

PR #1185, opened 2026-08-06
clamp snprintf result before fwrite in mtr_flush_with_state

aysha-afrah26:
MinitraceLogger passes node.name().c_str() to minitrace as the event name, and mtr_flush_with_state formats each event into a 1024 byte stack buffer before writing it out. The length handed to fwrite is snprintf's return value, which is the size the line would have needed rather than the size it actually wrote, so any event that overflows the buffer makes fwrite read past the end of linebuf. Node names come from the name attribute in the tree XML and nothing bounds their length, and mtr_flush() runs on every status transition, so a name of about a thousand characters is enough to reach it. The bytes read past the buffer are adjacent stack memory and they land verbatim in the trace file. I hit it under AddressSanitizer while looking at the logger tests: it reports a stack-buffer-overflow read of 1081 bytes out of the 1024 byte buffer at minitrace.cpp:356. Clamping len to what snprintf actually produced closes it, and treating a negative return as nothing-to-write also covers the MSVC _snprintf case where truncation reports -1 and fwrite would be called with (size_t)-1. I checked the rest of the tree for the same shape and this is the only place an snprintf return value is used as a length; the test added to gtest_loggers fails under ASan before the change and passes after.

### DIFF
--- a/3rdparty/minitrace/minitrace.cpp
+++ b/3rdparty/minitrace/minitrace.cpp
@@ -353,6 +353,14 @@ void mtr_flush_with_state(int is_last) {
 		len = snprintf(linebuf, ARRAY_SIZE(linebuf), "%s{\"cat\":\"%s\",\"pid\":%i,\"tid\":%i,\"ts\":%" PRId64 ",\"ph\":\"%c\",\"name\":\"%s\",\"args\":{%s}%s}",
 				first_line ? "" : ",\n",
 				cat, raw->pid, raw->tid, raw->ts - time_offset, raw->ph, raw->name, arg_buf, id_buf);
+		// snprintf returns the length the line would have needed, not what it
+		// wrote, and a negative value on error, so clamp before using it as a
+		// write length.
+		if (len < 0) {
+			len = 0;
+		} else if ((size_t)len >= sizeof(linebuf)) {
+			len = (int)sizeof(linebuf) - 1;
+		}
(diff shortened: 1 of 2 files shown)


aysha-afrah26:
any update?

fallenmi (approved):
Verified on exact head `30ee5b3447b137b4b2d207b4f773f5f81bd6e24d` with an independent ASan oracle compiled directly against `minitrace.cpp`. The same 1,200-byte event name makes exact base `c88a9f429a421b312599a07fa8902524b09bf90a` abort with a stack-buffer-overflow read of 1,280 bytes from the 1,024-byte `linebuf` at `mtr_flush_with_state()` / `fwrite`; exact head exits 0 and caps the trace output instead. The clamp also keeps a negative formatter return from becoming a huge unsigned write length. `git diff --check` passes and all 15 current upstream checks are green.

AI-assisted review performed with OpenAI Codex under `fallenmi` direction; the exact base/head oracle and live PR state were independently verified before submission.

facontidavide:
this is a third party library, we do not modify it here, but upstream first

aysha-afrah26:
Makes sense. Sent it upstream: hrydgard/minitrace#49, same clamp against their minitrace.c. Once it lands there I can rework this to just sync the vendored copy, or close it if you'd rather pick the fix up with the next minitrace update.

PR #1186, opened 2026-08-10
preserve integer width in JsonExporter::fromJson

aysha-afrah26:
ExportBlackboardToJSON writes int64_t and uint64_t entries out at full width, but fromJson read every JSON integer back with nlohmann's get<int>(), which is just a static_cast<int> of the stored 64-bit value with no range check. Anything outside the int32 range wraps silently, so a saved int64_t of 5000000000 comes back as 705032704 and 4294967296 comes back as 0, and the same happens to any value fed through ImportBlackboardFromJSON or ImportTreeFromJSON when restoring tree state. I hit it doing an export/import round-trip and noticing the restored counter no longer matched. The fix keeps the width during import: values that fit in int stay int, larger magnitudes stay uint64_t or int64_t to match what toJson already emits. Small integers are unaffected, so the common case behaves exactly as before. Added a regression test that imports the out-of-range values and checks the round-trip.

### DIFF
--- a/src/json_export.cpp
+++ b/src/json_export.cpp
@@ -1,5 +1,7 @@
 #include "behaviortree_cpp/json_export.h"
 
+#include <limits>
+
 namespace BT
 {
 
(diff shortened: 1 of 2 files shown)


aysha-afrah26:
any update?

fallenmi (approved):
Reviewed the scalar integer dispatch and regression coverage. Checking unsigned values before signed integers, preserving the existing `int` representation inside its range, and falling back to the matching 64-bit type outside that range is consistent with the exporter and avoids the previous silent truncation. The round-trip and compatibility cases cover the behavioral boundary, and the full CI matrix is green. This review was prepared with OpenAI Codex assistance under the account owner’s authorization.

Issue #934, opened 2025-02-14
Segmentation fault when substituting subtree


Isarm:
**Describe the bug**
Trying to add a substitution rule on a subtree results in a segmentation fault.  

**How to Reproduce**

```
TEST(MinimalExampleTest, minimalExampleTest) {
  BT::BehaviorTreeFactory factory;
  // Find all the XML files in a folder and register all of them.
  // We will use std::filesystem::directory_iterator
  using std::filesystem::directory_iterator;
  for (auto const& entry : directory_iterator(BT_TREES_PATH)) {
    if (entry.path().extension() == ".xml") {
      factory.registerBehaviorTreeFromFile(entry.path().string());
    }
  }
  BT::TestNodeConfig config;
  factory.addSubstitutionRule("child", config);

  BT::Tree tree = factory.createTree("Parent");
}
```

With these 2 trees inside BT_TREES_PATH:

```xml
<root BTCPP_format="4">
  <BehaviorTree ID="Parent" name="parent">
    <SubTree ID="ChildTree" name="child" />
  </BehaviorTree>
</root>
```
```xml
<root BTCPP_format="4">
  <BehaviorTree ID="Child">
    <AlwaysSuccess />
  </BehaviorTree>
</root>
```

PR #1193, opened 2026-09-01
validate substituted node type in instantiateTreeNode

aysha-afrah26:
Substitution rules replace a node by ID, but `instantiateTreeNode` never checks that the replacement has a compatible node type. The XML around the node was validated for its ORIGINAL type (mandatory ID, child count), so a compact leaf swapped for a structural node is mishandled at build time:

    <AlwaysSuccess name="action_A"/>  +  addSubstitutionRule("action_A", "SubTree")
    => Segmentation fault (exit 139)

The crash is `std::string(element->Attribute("ID"))` at xml_parsing.cpp:1095. A compact `<AlwaysSuccess/>` carries no ID, so `Attribute("ID")` is null and the string constructor calls `strlen(nullptr)`. The same setup with `"Inverter"` leaves the decorator `child_node_` null and segfaults on the first tick instead. Both are reachable from an untrusted `SubstitutionRules` JSON plus ordinary XML, before any node runs.

The fix rejects a substitution whose resulting type differs from the original unless the replacement is a leaf (Action or Condition). That keeps the supported cases working, same-type swaps and the `SubTree` to `TestNode` mocking path from #934, and turns both crashes into a clear `RuntimeError`. Added a regression test that segfaults before the change and passes after.

### DIFF
--- a/src/bt_factory.cpp
+++ b/src/bt_factory.cpp
@@ -365,6 +365,28 @@ std::unique_ptr<TreeNode> BehaviorTreeFactory::instantiateTreeNode(
     }
   }
 
+  // A substitution rule must not turn a node into a structurally incompatible
+  // type. The XML was validated (children count, mandatory ID) against the
+  // original node type, so replacing a leaf with a SubTree, Decorator or
+  // Control leaves the builder with a node whose child or "ID" attribute the
+  // XML never supplied. That later dereferences a null child pointer or builds
+  // a std::string from a null Attribute("ID"). Allow same-type swaps and swaps
+  // to a leaf (the common "replace with a mock action" case, including the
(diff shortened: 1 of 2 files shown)


aysha-afrah26:
Pushed a small follow-up: the new test was tripping MSVC C4834 (discarded nodiscard result), so it now uses the same `(void)` cast as the other factory tests.

The two red jobs on the first run were timing flakes in `RepeatTestAsync.RepeatTestAsync` and `CoroTest.sequence_child`. Neither goes through the factory, and both pass on the re-run. The windows Debug job has been sitting in the Build step for a while on this run (Release built the same sources fine), so that one looks like a runner stall and may need a re-run.

